PR #3, opened 2016-09-14
Create nextant.csr

daita:
my app certificate for https://github.com/daita/nextant/

@LukasReschke


LukasReschke:
Thanks! Your certificate can be found at https://github.com/nextcloud/app-certificate-requests/blob/master/nextant/nextant.crt

PR #12, opened 2016-09-19
Add sms csr for sms app

nerzhul:
Hi @LukasReschke ,
can you accept this request, i need to publish my app


LukasReschke:
Can you link to your app? https://github.com/nerzhul/ocsms seems to have the app id "ocsms" instead of "sms".

nerzhul:
yes this is the app... the certificate should match the app name ? It's owncloudSMS the old name i cannot change to keep backward compat, should i keep ocsms in this CSR too ?

LukasReschke:
The CN has to match the app id: https://github.com/nerzhul/ocsms/blob/6c906d498f01b94dad5110999cbcc39b98377446/appinfo/info.xml#L3

nerzhul:
okay i'm re-doing it, closing this to win some time

PR #14, opened 2016-09-21
Request keeweb

jhass:
/cc @LukasReschke


LukasReschke:
Thanks. Your certificate is at https://github.com/nextcloud/app-certificate-requests/blob/master/keeweb/keeweb.crt

PR #23, opened 2016-10-16
Update csr

brantje:


BernhardPosselt:
Why another CSR?

LukasReschke:
https://github.com/nextcloud/app-certificate-requests/blob/master/passman/passman.crt

PR #28, opened 2016-11-23
Create spreed.csr

Ivansss:
@LukasReschke


LukasReschke:
https://github.com/nextcloud/app-certificate-requests/blob/master/spreed/spreed.crt

PR #33, opened 2016-12-07
CSR for PreviewGenerator

rullzer:
See https://github.com/rullzer/previewgenerator

CC: @LukasReschke


LukasReschke:
https://github.com/nextcloud/app-certificate-requests/blob/109ff705dc37c545f4b32ec8f837a030d6ddeb7e/previewgenerator/previewgenerator.crt

PR #46, opened 2017-02-10
Create files_inotify.csr

icewind1991:
cc @LukasReschke


LukasReschke:
https://github.com/nextcloud/app-certificate-requests/blob/master/files_inotify/files_inotify.crt

PR #61, opened 2017-04-26
Create groupfolders.csr

icewind1991:
cc @LukasReschke


LukasReschke:
Is that an official app? Then I'd prefer it to be bundled from the Nextcloud account with a key from the packaging system.

icewind1991:
Ok

PR #79, opened 2017-07-26
Add drop_account csr

tcitworld:
`drop_account` is an app to allow users to remove their account. Below are some informations for the curious.

### Background
We (will) use it at @framasoft for our services Framadrive and Framagenda both based on Nextcloud. Framadrive was limited to ~5000 accounts but Framagenda is opened to all through the registration app.
We regularly ask users to close their accounts to free some space and make room for others, but they needed to contact us in order to do so. It's now simpler for them to do it directly from Nextcloud.

### Future evolutions
What I plan : 
* Send an email to the user when the account has been deleted
* Admin option to close the account only through a token validation sent through email
* Admin option to deactivate the accounts, not delete them (they should be deleted after a while by a cron job)

### Sources
https://framagit.org/tcit/drop_user


LukasReschke:
https://github.com/nextcloud/app-certificate-requests/blob/master/drop_account/drop_account.crt :)

PR #84, opened 2017-08-06
Create  files_clipboard.csr

leizh:
@LukasReschke


LukasReschke:
https://github.com/nextcloud/app-certificate-requests/blob/master/files_clipboard/files_clipboard.crt

PR #90, opened 2017-08-27
PhoneTrack csr

eneiluj:
@LukasReschke : A csr for my new app


eneiluj:
Here is the repo : https://gitlab.com/eneiluj/phonetrack-oc

LukasReschke:
There we go: https://github.com/nextcloud/app-certificate-requests/blob/master/phonetrack/phonetrack.crt :)

PR #96, opened 2017-10-06
fullnextsearch (core)

daita:


LukasReschke:
https://github.com/nextcloud/app-certificate-requests/blob/master/fullnextsearch/fullnextsearch.crt

PR #100, opened 2017-10-21
Create mindmaps.csr

splitt3r:
App: https://github.com/splitt3r/mindmaps

PR #100 🎈


LukasReschke:
There we go: https://github.com/nextcloud/app-certificate-requests/blob/master/mindmaps/mindmaps.crt

PR #108, opened 2017-11-16
create nextcloud.csr

hanysayed:
Thanks in advanced for sending the signing certificate


nickvergessen:
The filename should be your appid and be stored in a folder which is also the appid

PR #118, opened 2018-01-04
search.csr

daita:
In case https://github.com/nextcloud/fullnextsearch/issues/225 stays with 'search'


MorrisJobke:
We should use "Full text search"

PR #120, opened 2018-01-04
fulltextsearch

daita:


MorrisJobke:
Why are there 3 CSRs in here? fulltextsearch should be the only one, right?

daita:
Cause right now, there is [3 apps](https://github.com/nextcloud/fullnextsearch/wiki)

the Core, a Provider to index the files and a Platform to perform searchs

PR #123, opened 2018-01-10
Create w2g2.csr

Emi94:
This is a fork for the working2gether app. As there is no response from the maintainer, we (@newroco) decided to fork it and fix it ourselves.
https://github.com/newroco/W2G2
@MorrisJobke


MorrisJobke:
828c625

PR #132, opened 2018-02-03
Taking over ownership of the contacts app

Henni:
Signed-off-by: Hendrik Leppelsack <hendrik@leppelsack.de>


MorrisJobke:
Could you not send over the keys? Otherwise we need to revoke the existing certificate. Also I could simply add you as co-maintainer and then both of you (existing owner and you) can upload packages to the app store.

Henni:
solved through key transfer

PR #135, opened 2018-02-26
recommendation_assistant.csr

doganoo:
@MorrisJobke as discussed in https://github.com/nextcloud-gmbh/recommendation_assistant/issues/1

can I make changes after code signing or does each change require its own another code-sign?


doganoo:
fixed typo in CN

PR #137, opened 2018-03-15
Signing Certificate request for twofactory_yubikey

jaark:
2FA implementation for Yubikey HOTP device
@MorrisJobke

https://github.com/jaark/twofactor_yubikey


MorrisJobke:
cc @ChristophWurst Maybe join forces on this app ;)

MorrisJobke:
f052e50

jaark:
Thanks. 
The OTP and U2F modes of the Yubikey operate almost as though it is a totally different device. Although all Yubikeys support OTP, only newer ones support U2F and OTP does not need special browser support.

Though it might be possible to implement OTP as a fall-back mode for U2F authentication though with the differences, I don't think that this would be practical (though Christoph might know better than I)

That said, I always welcome peer review of code ;)

PR #155, opened 2018-06-05
Create auto_mail_accounts.csr

PanCakeConnaisseur:
https://github.com/PanCakeConnaisseur/auto_mail_accounts

@MorrisJobke


MorrisJobke:
@ChristophWurst @jancborchardt Something related to the mail app ;) You maybe want to join forces ;)

jancborchardt:
Seems like this is something which should be in the Mail app directly instead of a separate app? @PanCakeConnaisseur @ChristophWurst @Gomez what do you think?

PR #157, opened 2018-06-18
Create ransomware_detection.csr

ilovemilk:
cc @MorrisJobke 
https://github.com/ilovemilk/ransomware_detection


MorrisJobke:
Is there a way to combine this into the ransomware app, that already exists? Join forces helps to keep the app maintained: https://github.com/nextcloud/ransomware_protection

cc @nickvergessen for this. 

I would keep this open until Wednesday (when @nickvergessen has time to have a look at this) if this is okay for you?

ilovemilk:
Sure, thats okay with me!

These both apps are basically different: Ransomware protection uses ransomware-specific file extensions to block synchronization requests in real-time.
In contrast my application doesn't block any ransomware activities. It utilizes the file versioning of Nextcloud to easily recover from a ransomware attack by offering a recovery page,  with the color-coded classification of the file synchronization activities, for the users. Thus it works more generic and doesn't require maintainance of the dataset of ransomware-specific properties.

A possible option would be to use my application as fallback, if the blocking of ransomware files fails.

nickvergessen:
Yeah sounds good but we should not mix them I think.

MorrisJobke:
a79be50

PR #158, opened 2018-06-19
Create twofactor_sms.csr

ChristophWurst:
For https://github.com/ChristophWurst/twofactor_sms

cc @MorrisJobke


ChristophWurst:
Wait … let me actually rename the app before I request a cert

PR #161, opened 2018-06-30
Create longfilename.csr

pellaeon:
https://github.com/pellaeon/longfilename


pellaeon:
@MorrisJobke

MorrisJobke:
> https://github.com/pellaeon/longfilename

Shouldn't this be fixed in the server itself?

cc @nextcloud/designers for this one here. I guess we should fix it properly instead of relying on an app to fix broken UI.

skjnldsv:
Agreed

jancborchardt:
Yes, this is something we should deal with in the server. But: Allowing 2 lines is fine, but where do you set the limit?

skjnldsv:
@jancborchardt time to bring back marquee!!! :D

MorrisJobke:
The current UI shows the full file name in a tool tip:

bildschirmfoto 2018-10-11 um 15 06 16

I will close this request for now as this app is then not needed, right @pellaeon?

pellaeon:
Sorry, but I thought the point of having the sidebar is to display "detailed" information about the file?

I understand that you don't want the long filename to clutter up the interface, but isn't that the reason why we have the sidebar?

With the tooltip, the filename is displayed twice.

1. We don't want the lone filename to clutter up the file list, so we put it in the sidebar.
2. We don't want the lone filename to clutter up the sidebar, so we put the long filename in the tooltip.
I don't think it makes sense.

And I also fully understand the nextcloud team have certain design decisions and opinions and I respect that, but maybe for this one, it could be a user choice, by allowing this app in the appstore? the user can decide what design makes sense for themself.

MorrisJobke:
> And I also fully understand the nextcloud team have certain design decisions and opinions and I respect that, but maybe for this one, it could be a user choice, by allowing this app in the appstore? the user can decide what design makes sense for themself.

I fully understand the problem here, but it shouldn’t be a „let the users decide“ just because our design is not working here. We then should adjust the design and not give the users config options.

Could you open a bug report in the server repo?

PR #182, opened 2018-09-26
Create APP_ID.csr

elosapio:


MorrisJobke:
it looks like you did not replaced the placeholder `APP_ID` with the app ID of your actual app. Be also aware that this is only needed if you want to upload your app to https://apps.nextcloud.com. I will close this PR and you could send in a new one with the correct app ID if needed.

PR #183, opened 2018-09-26
Create tcboxbeta.csr

elosapio:
@MorrisJobke


MorrisJobke:
Is there a reason why you need actually two apps? (see #184)

Do you also have a repo where I can check the app ID to be the correct one?

elosapio:
no, sorry

PR #187, opened 2018-09-27
Create hernosapp.key

Hernos:
@MorrisJobke


nickvergessen:
You uploaded your private key. Please generate a new one. You should only upload the .csr file

PR #193, opened 2018-10-07
Update linkshareex.csr

yaoteng:
@MorrisJobke


MorrisJobke:
I will revert the old cert. That makes it impossible to install apps with the old signature. Just that you know that a re-release of all version of the app that are still supported is needed.

MorrisJobke:
The cert is in 362b472

PR #207, opened 2018-12-17
Certificate for files_readmemd

mamatt:
Hi NC team 

Certificate request for files_readmeMD
-> https://gitlab.univ-nantes.fr/uncloud/files_readmemd

@MorrisJobke


mamatt:
You were right it was missing some documentation/explanation  ;)
I've updated the readme file !

jancborchardt:
@mamatt thanks a lot, cool! :)

One possible issue I could see is that "MarkDown editor need to be installed". We try not to have cross-app dependencies. But needs input from @rullzer @MorrisJobke how we handle this.

mamatt:
The apps check that markdown editor is installed.
=> https://gitlab.univ-nantes.fr/uncloud/files_readmemd/blob/master/js/script.js#L30

rullzer:
> @mamatt thanks a lot, cool! :)
> 
> One possible issue I could see is that "MarkDown editor need to be installed". We try not to have cross-app dependencies. But needs input from @rullzer @MorrisJobke how we handle this.

Fine I'd say. We do try to avoid it but here it is clear. And they check for it. :)

mamatt:
For the next version, I'll try to switch from MD to raw sanitized txt if the MD editor app is not present.

PR #208, opened 2018-12-21
certificate for user_external

violoncelloCH:
certificate for the user_external app to publish it to the appstore and make it independent from server releases
cc @nextcloud/user_external @jancborchardt 
for nextcloud/user_external#22 

Signed-off-by: Jonas Sulzer <jonas@violoncello.ch>


jancborchardt:
@MariusBluem mentioned that the app is marked as official/shipped and should maybe have "Nextcloud" as a publisher?

@rullzer @MorrisJobke what's your call?

rullzer:
> @MariusBluem mentioned that the app is marked as official/shipped and should maybe have "Nextcloud" as a publisher?
> 
> @rullzer @MorrisJobke what's your call?

Doesn't matter much IMO.

rullzer:
So. I'm a bit hessitant. In the past we have had issues with apps that were both shipped and in the appstore when upgrading.

MariusBluem:
We could remove the shipped-flag and remove it from the build script with 15.0.1. Minimal compatible version would be 15, all other old releases will not be affected.
Its not working with 15.0.0 at all in the current release so we would at least not break something ...

violoncelloCH:
> @MariusBluem mentioned that the app is marked as official/shipped and should maybe have "Nextcloud" as a publisher?

how could I achieve this? can I just set "Nextcloud" as `<author>`?

I already thought I'll have to remove the shipped-flag. see PR nextcloud/user_external#26

MorrisJobke:
> how could I achieve this? can I just set "Nextcloud" as `<author>`?

That is only defined in the app store.

PR #216, opened 2019-01-29
Update passman csr

animalillo:
@MorrisJobke please check this new csr for the passman app as the old certificate seems to have expired
https://github.com/nextcloud/passman


MorrisJobke:
@newhinton @brantje Is this fine with you?

MorrisJobke:
The cert in here is valid till 2027:

```
        Validity
            Not Before: Oct 19 19:35:12 2016 GMT
            Not After : Jan 25 19:35:12 2027 GMT
        Subject: CN=passman
```

animalillo:
I requested this because @brantje said the problem with the app was with the current certificate being expired, but we can check other possible causes if that's not the issue

brantje:
@MorrisJobke No need to ask @newhinton ;), i'm still main author.
I've asked @animalillo to update the cert.

PR #219, opened 2019-02-05
add payback.csr

eneiluj:
Hi there!

Here is my certificate request for a new app named [Payback](https://gitlab.com/eneiluj/payback-nc/). It's a shared budget manager like [IHateMoney](https://github.com/spiral-project/ihatemoney/).

Project is still young but all base features work. Style guidelines are almost fully respected. Code check passes.


MorrisJobke:
Hey @eneiluj Thanks for this - I'm a bit unsure about the name, because at least in Germany it's a name of a big bonus points company: https://de.wikipedia.org/wiki/Payback_(Bonusprogramm) and it might be that we or you get requests to name it differently. Thus I would bring this up before issuing a cert and then need to redo it in another way anyways.

eneiluj:
Ow too bad. I thought there was no restriction to use words from the dictionary...

I already eliminated "KillBill" because of potential copyright mess :smile:.

The only other names i came up with are too long or sound bad in my opinion. Things like "ShareTheBill", "GroupSpend"...

What about "CoSpend" ? There already is a proprietary Android app named CoSpender.

I didn't think finding a name for such a simple app purpose was going to be that difficult.

MorrisJobke:
> Ow too bad. I thought there was no restriction to use words from the dictionary...
> 
> I already eliminated "KillBill" because of potential copyright mess 😄.
> 
> The only other names i came up with are too long or sound bad in my opinion. Things like "ShareTheBill", "GroupSpend"...
> 
> What about "CoSpend" ? There already is a proprietary Android app named CoSpender.
> 
> I didn't think finding a name for such a simple app purpose was going to be that difficult.

This was just my thought. I could sign it and then we can look into how this evolves and if it matters at all, but as this is a quite big player I don't think that it is too unlikely.

eneiluj:
No i guess you're right. Let's delete this PR and i'll come back with a better name.

Do you think "CoSpend" is right even if it's very close to CoSpender?

MorrisJobke:
> Do you think "CoSpend" is right even if it's very close to CoSpender?

Should be fine, because it is also another platform, but IANAL.

eneiluj:
Ok i'll think it through. Thanks for your answer!

rullzer:
I'd just do it. Worst comes to worst in the future it needs a rename.

eneiluj:
:smile: I'm okay with Payback obviously. Up to @MorrisJobke i guess.

MorrisJobke:
f394c36

eneiluj:
Thanks a lot to you both!

nickvergessen:
I like "share the bill", app images being an angry bill:


---

Maybe grouptabs has no copyright on their name? https://grouptabs.net/  cc @jancborchardt 

---

anyway, I would not have gone with payback, but we will see.

eneiluj:
Yep i like this one too but thought about it and it's not simple to pronounce :wink:

jancborchardt:
Payback is definitely an issue – as said it’s internationally recognized as this coupon program. Also in English it has negative connotations like "it’s payback time".

CoSpend is better, or Billsplit, or Sharecost or "Shared costs" or or or. (No, it’s not fine to use [Grouptabs](https://grouptabs.net/) as the name. ;)

eneiluj:
Thanks @jancborchardt for your remarks.

I liked the ambiguity of the word "payback". It sounds like a mafia movie title (and there is one with Mel Gibson :smile:)

I'm fine with CoSpend. I'll wait a bit just in case someone suggests something else.

MorrisJobke:
I revoked this.

MorrisJobke:
See #223 for the new app.

MorrisJobke:
566439d

PR #222, opened 2019-02-12
Create hsts.csr

sualko:
I maintain a Nextcloud instance on a webspace without a possibility to add the HSTS header via a configuration file, so I created an app which does it. Link to the repo: https://github.com/sualko/cloud_hsts

@MorrisJobke


MorrisJobke:
e6916ead0a460560a05361199af680d9edd1f287

PR #225, opened 2019-02-22
Create secsignid.csr

BjoernSecSign:
@MorrisJobke


BjoernSecSign:
The source code can be found [here](https://github.com/SecSign/SecSign-ID-Nextcloud-Plugin)

PR #241, opened 2019-05-17
Update cookbook.csr

mrzapp:
I have lost my original signing key and need to update the download URL, sorry for the inconvenience


MorrisJobke:
b967810

PR #249, opened 2019-06-27
Add CSR for printing app

e-alfred:
Hello,

I created a printing app for Nextcloud that allows to print directly to a CUPS/LPR-connected printer.

Here is the repository: https://github.com/e-alfred/nextcloud-printer

Maybe @jospoortvliet somebody at Nextcloud is interested in having a complete document management solution that allows both scanning (there is an app for that already on the App Store from me) and printing directly from Nextcloud?


MorrisJobke:
3858da8

PR #256, opened 2019-07-29
Request cert for user_retention

nickvergessen:
https://github.com/nextcloud/user_retention/


MorrisJobke:
@nickvergessen There was a problem on our side when the certificates were generated. There are a few certificates that have the same serial number. For now this is nothing to worry about, but once one certificate gets revoked all certificates with the same serial number won't work anymore. Thus I would ask you to take this new certificate for your app: 742730e. You key is still the same and the old certificate still works, but it might be that due to a revocation in the future the current one might be revoked as well. So please use the new one. Sorry for the inconvenience.

PR #257, opened 2019-07-29
Create files_audio_converter.csr

prsnbrg:
@MorrisJobke

[https://github.com/prsnbrg/files_audio_converter](https://github.com/prsnbrg/files_audio_converter)

thanks


blizzz:
https://github.com/nextcloud/app-certificate-requests/commit/c203fa25ac77357dd79600936fc1b720d794e195

Please excuse the delay

MorrisJobke:
@prsnbrg There was a problem on our side when the certificates were generated. There are a few certificates that have the same serial number. For now this is nothing to worry about, but once one certificate gets revoked all certificates with the same serial number won't work anymore. Thus I would ask you to take this new certificate for your app: a69a6bf. You key is still the same and the old certificate still works, but it might be that due to a revocation in the future the current one might be revoked as well. So please use the new one. Sorry for the inconvenience.

PR #258, opened 2019-07-30
Add sharingpath.csr

rookie0:
@MorrisJobke Tks.
https://github.com/rookie0/nextcloud-sharing-path


blizzz:
Since paths are not necessarily stable (the links breaks as soon as you move a file), do you perhaps see a potential overlap with https://apps.nextcloud.com/apps/sharerenamer ?

rookie0:
It's not overlap with sharerenamer, it's designed to as CDN origin storage，so the links breaks is acceptable.
The biggest problem is the potential security risk: links could be guessed(similar sharerenamer 😂),  but still acceptable for a CDN origin.

blizzz:
@rookie0 1af6be905ff3079fd6658dfac5ef0d7a6bf78b26

MorrisJobke:
@rookie0 There was a problem on our side when the certificates were generated. There are a few certificates that have the same serial number. For now this is nothing to worry about, but once one certificate gets revoked all certificates with the same serial number won't work anymore. Thus I would ask you to take this new certificate for your app: 538c042. You key is still the same and the old certificate still works, but it might be that due to a revocation in the future the current one might be revoked as well. So please use the new one. Sorry for the inconvenience.

PR #260, opened 2019-08-14
request Customized Emails SSL signing certificate

compagnon:
thanks you @MorrisJobke for Customized Emails App
[source code of my apps](https://github.com/compagnon/nextcloud-custom_apps)


juliushaertl:
When looking at the app it seems that it provides a rather company specific email template. Do you want to release that app to the app store or just use it internally? You don't need a certificate if you just want to run the app in your own companies instance.

compagnon:
> When looking at the app it seems that it provides a rather company specific email template. Do you want to release that app to the app store or just use it internally? You don't need a certificate if you just want to run the app in your own companies instance.

Objective is to make provide a framework to make possible a company specific emails , with HTML files inside the Themes directory. By default this app will provide HTML file of the NextCLoud standard email , instead of the hardcoded string inside the EmailTemplate class

blizzz:
88ba902eec071b376a2e96435ea592d683336cc4

MorrisJobke:
@compagnon There was a problem on our side when the certificates were generated. There are a few certificates that have the same serial number. For now this is nothing to worry about, but once one certificate gets revoked all certificates with the same serial number won't work anymore. Thus I would ask you to take this new certificate for your app: dbedd24. You key is still the same and the old certificate still works, but it might be that due to a revocation in the future the current one might be revoked as well. So please use the new one. Sorry for the inconvenience.

PR #261, opened 2019-08-21
push.csr

daita:
ex-stratos :~(


blizzz:
Repo: https://github.com/nextcloud/push

3787f0a8c84fa9ab8dc069ed77e84fbcd4cc2df7

MorrisJobke:
@daita There was a problem on our side when the certificates were generated. There are a few certificates that have the same serial number. For now this is nothing to worry about, but once one certificate gets revoked all certificates with the same serial number won't work anymore. Thus I would ask you to take this new certificate for your app: 300b14d. You key is still the same and the old certificate still works, but it might be that due to a revocation in the future the current one might be revoked as well. So please use the new one. Sorry for the inconvenience.

PR #262, opened 2019-09-18
Create pride.csr

skjnldsv:


rullzer:
Nice app.
You should be proud

blizzz:
repo url?

nickvergessen:
https://github.com/skjnldsv/pride

MorrisJobke:
@skjnldsv There was a problem on our side when the certificates were generated. There are a few certificates that have the same serial number. For now this is nothing to worry about right now, but once one certificate gets revoked all certificates with the same serial number won't work anymore. Thus I would ask you to take this new certificate for your app: cc5fc6c. You key is still the same and the old certificate still works, but it might be that due to a revocation in the future the current one might be revoked as well. So please use the new one. Sorry for the inconvenience.

PR #263, opened 2019-09-18
Create comicsans.csr

georgehrke:
@rullzer


nickvergessen:
Comic sans is actually a registered trademark, so I would like to not allow that as an appid

georgehrke:
It is? oO 

Ok, in that case maybe not put it into the app store.

blizzz:
:scream:

PR #264, opened 2019-09-18
Create calendar_resource_management.csr

georgehrke:
@rullzer


blizzz:
repo url?

blizzz:
certs/calendar_resource_management.crt

blizzz:
86b3cf5d89ecf9cb08345c682e9b8136ab32b045

MorrisJobke:
@georgehrke There was a problem on our side when the certificates were generated. There are a few certificates that have the same serial number. For now this is nothing to worry about, but once one certificate gets revoked all certificates with the same serial number won't work anymore. Thus I would ask you to take this new certificate for your app: 300b14d. You key is still the same and the old certificate still works, but it might be that due to a revocation in the future the current one might be revoked as well. So please use the new one. Sorry for the inconvenience.

PR #267, opened 2019-09-24
add mindmap_app.csr

fedwiiix:


blizzz:
@fedwiiix could you link to the repo, please?

blizzz:
Also, I think there are two mind mapping apps available right now. Have you considered joining forces?

fedwiiix:
My link is https://github.com/fedwiiix/mindmap_app

"joining forces" -> yes, of course, but the map representation are very different ...

blizzz:
alright, certificate is created: ba5e66dd28c50f48c06bb70cb07b9fc1131eec98

fedwiiix:
Hello, I do not know why but I no longer have the good csr, I changed the csr file on the git. Is the crt still good ?

when i test to add app in https://apps.nextcloud.com/developer/apps/new i have this error : 
`Signature is invalid: [('rsa routines', 'RSA_padding_check_PKCS1_type_1', 'invalid padding'), ('rsa routines', 'rsa_ossl_public_decrypt', 'padding check failed')]`

blizzz:
What's important now are the certificate (crt) file and the key file. Or those still the same?

MorrisJobke:
@fedwiiix There was a problem on our side when the certificates were generated. There are a few certificates that have the same serial number. For the Mindmaps_app the cert was revoked (not the Mindmaps app one but another one with the same serial number). Thus I would ask you to take this new certificate for your app: 9a4307a. You key is still the same. So please use the new one. Sorry for the inconvenience.

PR #270, opened 2019-10-17
Add CSR for OIDC Login app

pulsejet:
(Alpha-ready) OIDC login aims to be a very simplistic way to login to Nextcloud with an OpenID Connect provider for new or existing accounts. https://github.com/pulsejet/nextcloud-single-openid-connect


pulsejet:
@MorrisJobke

pulsejet:
@blizzz anything on this?

blizzz:
Thanks for your patience :)

blizzz:
5c1967498d3a28df2e1f067b1c4953f8d97efcf4

pulsejet:
Thank you @blizzz !

MorrisJobke:
@pulsejet There was a problem on our side when the certificates were generated. There are a few certificates that have the same serial number. For now this is nothing to worry about, but once one certificate gets revoked all certificates with the same serial number won't work anymore. Thus I would ask you to take this new certificate for your app: ac2e425. You key is still the same and the old certificate still works, but it might be that due to a revocation in the future the current one might be revoked as well. So please use the new one. Sorry for the inconvenience.

pulsejet:
Thanks for the heads up @MorrisJobke. I've uploaded a new release with the new certificate =)

PR #271, opened 2019-10-27
wopi app key

xklonx:
https://github.com/roscomputing/nextcloudwopi
@MorrisJobke


blizzz:
d0a27536e030f2a7a61d4dc7816178098a12cfb3

MorrisJobke:
@xklonx There was a problem on our side when the certificates were generated. There are a few certificates that have the same serial number. For now this is nothing to worry about, but once one certificate gets revoked all certificates with the same serial number won't work anymore. Thus I would ask you to take this new certificate for your app: There was a problem on our side when the certificates were generated. There are a few certificates that have the same serial number. For now this is nothing to worry about, but once one certificate gets revoked all certificates with the same serial number won't work anymore. Thus I would ask you to take this new certificate for your app: ENTERSHASUMHERE. You key is still the same and the old certificate still works, but it might be that due to a revocation in the future the current one might be revoked as well. So please use the new one. Sorry for the inconvenience.. You key is still the same and the old certificate still works, but it might be that due to a revocation in the future the current one might be revoked as well. So please use the new one. Sorry for the inconvenience.

MorrisJobke:
@xklonx There was a problem on our side when the certificates were generated. There are a few certificates that have the same serial number. For now this is nothing to worry about, but once one certificate gets revoked all certificates with the same serial number won't work anymore. Thus I would ask you to take this new certificate for your app: 2fceef8. You key is still the same and the old certificate still works, but it might be that due to a revocation in the future the current one might be revoked as well. So please use the new one. Sorry for the inconvenience.

xklonx:
@MorrisJobke Im trying to publish new release and getting `Certificate is invalid: [23, 0, 'certificate revoked']`. What must i do now, update app certificate in register app form?

nickvergessen:
https://github.com/nextcloud/app-certificate-requests/pull/287#issuecomment-644279715

nickvergessen:
Yeah, you need to repost your information at https://apps.nextcloud.com/developer/apps/new

PR #273, opened 2019-11-11
files_lock

daita:


blizzz:
647f18968d8be7ca9b34866f72045d51ef9048d2

MorrisJobke:
@daita There was a problem on our side when the certificates were generated. There are a few certificates that have the same serial number. For now this is nothing to worry about, but once one certificate gets revoked all certificates with the same serial number won't work anymore. Thus I would ask you to take this new certificate for your app: 58bca52. You key is still the same and the old certificate still works, but it might be that due to a revocation in the future the current one might be revoked as well. So please use the new one. Sorry for the inconvenience.

PR #94, opened 2017-10-05
Quicknotes and facerecognition..

matiasdelellis:
Quicknotes is a old app that already had published in owncloud...
 * https://github.com/matiasdelellis/quicknotes/
 * https://apps.owncloud.com/content/show.php/Quicknotes?content=174716
 
And facerecognition that is a new application that I am working and it implement a basic facial recognition system for nextcloud.
 * https://github.com/matiasdelellis/facerecognition

@LukasReschke
Thanks..

PR #274, opened 2019-11-14
Facerecognition

matiasdelellis:
Hi @MorrisJobke 

Due to an error in the development server I lost the certificate, although I never got to publish with it.
This PR replaces the old certificate with a new one.

Original PR: 
 * https://github.com/nextcloud/app-certificate-requests/pull/94

App source: 
 * https://github.com/matiasdelellis/facerecognition


rullzer:
I'll take care of this

rullzer:
See https://github.com/nextcloud/app-certificate-requests/blob/master/facerecognition/facerecognition.crt

matiasdelellis:
Thank you very much!!. :smiley:

MorrisJobke:
@matiasdelellis There was a problem on our side when the certificates were generated. There are a few certificates that have the same serial number. For now this is nothing to worry about, but once one certificate gets revoked all certificates with the same serial number won't work anymore. Thus I would ask you to take this new certificate for your app: 04154c8. You key is still the same and the old certificate still works, but it might be that due to a revocation in the future the current one might be revoked as well. So please use the new one. Sorry for the inconvenience.

matiasdelellis:
Ohh.. Ok .. Thank you very much.. :wink:

matiasdelellis:
@MorrisJobke 
Please, I am having trouble publishing the application.

> You key is still the same and the old certificate still works, but it might be that due to a revocation in the future the current one might be revoked as well. So please use the new one. Sorry for the inconvenience.

When I upload a release, just get:

> Certificate is invalid: [23, 0, 'certificate revoked']

From my part? What do I have to do?.

matiasdelellis:
Hi, Sorry for the spam. :wink: 

Finally register the app again, and it work. Thanks.

MorrisJobke:
> Hi, Sorry for the spam. 😉
> 
> Finally register the app again, and it work. Thanks.

Don't worry :) We made some mistakes and you needed to take action. Good that you figured it out 👍

PR #276, opened 2019-11-27
Create nextbackup.csr

pbek:
repo: https://github.com/pbek/nextbackup

I have to rename ownbackup to nextbackup.
I got a letter from https://www.ownbackup.com/. :grimacing:


pbek:
@MorrisJobke, thank you

blizzz:
@pbek here you go d24d0bb64697afd1ca431698db64af54da008840

blizzz:
@rullzer i think you can do the revoke dance around the old ownbackup cert then

pbek:
@blizzz thank you very much!

pbek:
Is there a way I can "unpublish", disable or archive the ownBackup store page?

blizzz:
@pbek looks like i can only totally delete it…

blizzz:
or i could possibly rename it, however you got already the new one up.

pbek:
@blizzz, renaming sounds great too, in that case you will delete NextBackup, rename OwnBackup to NextBackup and I will simply publish the latest release again, right? If yes, please go for it.

blizzz:
@pbek I did the changes. Alas, the admin panel behaves strange, it  appears both app ids are present now :-/ The search shows ownbackup only. Please try to publish the release… if this fails, it indeed has to be the new entry I am afraid.

I am off the next days, in doubt ping @rullzer for so long.

pbek:
@blizzz thank you very much, I'll try on Saturday!

pbek:
@blizzz, @rullzer Seems like I can't push a new release now.

When using the ownbackup cert I get a `App id nextbackup does not match cert CN ownbackup`
When using the nextbackup cert I get a `Signature is invalid: [('rsa routines', 'RSA_padding_check_PKCS1_type_1', 'invalid padding'), ('rsa routines', 'rsa_ossl_public_decrypt', 'padding check failed')]`.

I also tried re-packaging, uploading and signing again. Same result.
So I guess you will have to remove the current NextBackup app again so I can push a fresh one.

pbek:
@blizzz, @rullzer any news? 😬

rullzer:
It seems the certificate doesn't match let me have a look

rullzer:
Yes. Please try again. If it still fails I will just ermove ownBackup

pbek:
@rullzer, the upload worked and https://apps.nextcloud.com/apps/nextbackup has the current release, but it seems to be a fresh application, no comments of ownbackup are present.

https://apps.nextcloud.com/apps/ownbackup still exists with comments and so on.

pbek:
Any news on the removal? :grin:

MorrisJobke:
@pbek There was a problem on our side when the certificates were generated. There are a few certificates that have the same serial number. For now this is nothing to worry about, but once one certificate gets revoked all certificates with the same serial number won't work anymore. Thus I would ask you to take this new certificate for your app: 8e060f7. You key is still the same and the old certificate still works, but it might be that due to a revocation in the future the current one might be revoked as well. So please use the new one. Sorry for the inconvenience.

pbek:
Thank you @MorrisJobke, I will then from now on (next release of NextBackup) use the new `nextbackup.crt`.

MorrisJobke:
> Thank you @MorrisJobke, I will then from now on (next release of NextBackup) use the new `nextbackup.crt`.

@pbek Perfect :)

PR #277, opened 2019-12-02
EmlViewer.csr

a11exandru:
Add EmlViewer.csr


a11exandru:
@MorrisJobke

rullzer:
There you go

https://github.com/nextcloud/app-certificate-requests/blob/master/emlviwer/EmlViewer.crt

a11exandru:
Thanks, but i can't upload the app release because i generate the request certificate for "EmlViewer.csr" instead of "emlviewer.csr", and if i try to upload this i get the error of "App emlviewer does not exist, you need to registerit first".
@rullzer It's ok to regenerate the certificate? Or exists other solution to solve that? 
And i'm really sorry for that!

a11exandru:
If the best solution is to generate a new certificate, there is the new PR: https://github.com/nextcloud/app-certificate-requests/pull/280 .

rullzer:
Yeah a new certificate. All lowercase.

a11exandru:
Ok, here is the link: https://github.com/nextcloud/app-certificate-requests/pull/280

PR #278, opened 2019-12-04
Changed ldapcontacts.csr due to loss

alexhor:
https://github.com/KircheNeuenburg/ldapcontacts


alexhor:
@MorrisJobke

rullzer:
https://github.com/nextcloud/app-certificate-requests/blob/master/ldapcontacts/ldapcontacts.crt

MorrisJobke:
@alexhor There was a problem on our side when the certificates were generated. There are a few certificates that have the same serial number. For now this is nothing to worry about, but once one certificate gets revoked all certificates with the same serial number won't work anymore. Thus I would ask you to take this new certificate for your app: 7a8ff73. You key is still the same and the old certificate still works, but it might be that due to a revocation in the future the current one might be revoked as well. So please use the new one. Sorry for the inconvenience.

PR #281, opened 2019-12-13
Certificate for Joplin app

laurent22:
@MorrisJobke, this is the certificate for my Joplin Nextcloud app. Source code for it is there: https://github.com/laurent22/joplin-nextcloud/ Official website is there: https://joplinapp.org

If you need any additional information, please let me know.


blizzz:
@laurent22 here you go 7407c4d4e9a735fb7d48388d30afa8460dacedc4

laurent22:
Thanks @blizzz!

MorrisJobke:
@laurent22 There was a problem on our side when the certificates were generated. There are a few certificates that have the same serial number. For now this is nothing to worry about, but once one certificate gets revoked all certificates with the same serial number won't work anymore. Thus I would ask you to take this new certificate for your app: 424dc5b. You key is still the same and the old certificate still works, but it might be that due to a revocation in the future the current one might be revoked as well. So please use the new one. Sorry for the inconvenience.

PR #282, opened 2019-12-18
Request for files_texteditor

juliushaertl:
Let's do a release of files_texteditor in the app store since a lot of people still prefer a plain text editor or files_markdown for editing. https://github.com/nextcloud/files_texteditor/issues/185

cc @rullzer @blizzz


blizzz:
8a0ca6151dcbbde65a5795f19f6adfbd474fc86f

MorrisJobke:
@juliushaertl There was a problem on our side when the certificates were generated. There are a few certificates that have the same serial number. For now this is nothing to worry about, but once one certificate gets revoked all certificates with the same serial number won't work anymore. Thus I would ask you to take this new certificate for your app: 099dfc6. You key is still the same and the old certificate still works, but it might be that due to a revocation in the future the current one might be revoked as well. So please use the new one. Sorry for the inconvenience.

PR #60, opened 2017-04-12
Create spgverein.csr

schrieveslaach:
@LukasReschke Could you accept the [spgverein app](https://gitlab.com/schrieveslaach/nextcloud-spgverein-app) please? Thanks!

PR #283, opened 2019-12-28
Update spgverein.crt

schrieveslaach:
Somehow the verification of spgverein fails (see [here](https://gitlab.com/schrieveslaach/nextcloud-spgverein-app/issues/2)). As proposed [here](https://github.com/nextcloud/app-certificate-requests/pull/60#issuecomment-530276121) the simplest solution would be to recreate the certificate.


rullzer:
I'll take care of this tomorrow

rullzer:
See https://github.com/nextcloud/app-certificate-requests/blob/master/spgverein/spgverein.crt

mburnicki:
Hi, I just came across the message that the certificate of the app was invalid when I tried to install it on NC 17.0.2 from the app store.

Does the new cert need to be published or pushed to the store, or is the new cert still considered invalid?

schrieveslaach:
@mburnicki, I made a very stupid mistake. I lost my certificate key. I have to make a new PR and the I will upload a signed version of the plugin. @rullzer, I'm very sorry, about that.

PR #285, opened 2020-01-04
Create breezedark.csr

mwalbeck:
@MorrisJobke 

The source code is found here: https://github.com/mwalbeck/nextcloud-breeze-dark/tree/app


blizzz:
here you go 62771343933be4fbdbd5d0a19981cc73cfd0c29f :)

MorrisJobke:
@mwalbeck There was a problem on our side when the certificates were generated. There are a few certificates that have the same serial number. For now this is nothing to worry about, but once one certificate gets revoked all certificates with the same serial number won't work anymore. Thus I would ask you to take this new certificate for your app: e63acfb. You key is still the same and the old certificate still works, but it might be that due to a revocation in the future the current one might be revoked as well. So please use the new one. Sorry for the inconvenience.

mwalbeck:
@MorrisJobke No worries, thanks for the quick response on this. I'll be using the new cert moving forward.

PR #287, opened 2020-01-12
Create geoblocker.csr

HomeITAdmin:
@MorrisJobke

Source Code App: https://github.com/HomeITAdmin/nextcloud_geoblocker


blizzz:
here you go: 06857111e8b3773bd924d3df84e4c92b7029dacb

MorrisJobke:
@HomeITAdmin There was a problem on our side when the certificates were generated. There are a few certificates that have the same serial number. For now this is nothing to worry about, but once one certificate gets revoked all certificates with the same serial number won't work anymore. Thus I would ask you to take this new certificate for your app: 13c75df. You key is still the same and the old certificate still works, but it might be that due to a revocation in the future the current one might be revoked as well. So please use the new one. Sorry for the inconvenience.

HomeITAdmin:
Thanks, I will use the new one from now on.

HomeITAdmin:
@MorrisJobke I am trying to upload a new version of the geoblocker app, but keep getting the error message: Certificate is invalid: [23, 0, 'certificate revoked']
Could this be connected?

I am not sure, what is going wrong. For signing the app I am just using the private key, which didn't change, and not the public certificate. So there shouldn't be a problem. Do I miss something here?

nickvergessen:
I had the same issue with one of my apps. Debugged it with sysadmin for a while and in the end we decided to do a new cert instead. See #314

I guess you can just do the same thing

HomeITAdmin:
Thanks for the hint, this brought me to point that I had to register the App again with the new certificate. I forgot this. Now it is working with the provided certificate and the old key in this case.

nickvergessen:
Oh let me try that too

nickvergessen:
That worked for me too, thanks for the pointer!

PR #289, opened 2020-02-16
New app certificate request for spgverein

schrieveslaach:
As mentioned in #283 a lost my private key. Therefore, here is a new request.


schrieveslaach:
@rullzer, could you help me here?

MorrisJobke:
Let me revoke the old cert and create a new one.

MorrisJobke:
352bda6

MorrisJobke:
@schrieveslaach There was a problem on our side when the certificates were generated. There are a few certificates that have the same serial number. For now this is nothing to worry about, but once one certificate gets revoked all certificates with the same serial number won't work anymore. Thus I would ask you to take this new certificate for your app: b51f195. You key is still the same and the old certificate still works, but it might be that due to a revocation in the future the current one might be revoked as well. So please use the new one. Sorry for the inconvenience.

PR #293, opened 2020-03-20
Create epubreader.csr

e-alfred:
@MorrisJobke (still there?)

I am reviving the files_epubreader app: https://github.com/e-alfred/owncloud-apps/tree/master/files_reader

The original app is still listed in the App Store by @Yetangitu, but unmaintained for a long time now: https://apps.nextcloud.com/apps/files_reader


blizzz:
@e-alfred sorry for the delay! The certificate name should match the appid. Will you change it to epubreader, too? In the code it is still files_reader.

e-alfred:
Sorry, I hadn't finished polishing it, the appid should be fixed now.

MorrisJobke:
4c1ebf3

e-alfred:
Repository is now at https://github.com/e-alfred/epubreader

MorrisJobke:
@e-alfred There was a problem on our side when the certificates were generated. There are a few certificates that have the same serial number. For now this is nothing to worry about, but once one certificate gets revoked all certificates with the same serial number won't work anymore. Thus I would ask you to take this new certificate for your app: a23bf7d. You key is still the same and the old certificate still works, but it might be that due to a revocation in the future the current one might be revoked as well. So please use the new one. Sorry for the inconvenience.

PR #295, opened 2020-04-04
Create Sendent.csr

leonvandebroek:
As the creators of the Nextcloud Outlook Add-in we are looking to publish our server settings app to the App Store. As this requires a code signing certificate from Nextcloud. This is the pull request to get that certificate.
@MorrisJobke


blizzz:
@leonvandebroek would you have an URL to the code?

leonvandebroek:
The code is not public, is that a problem?

MorrisJobke:
In general we check against the appid in the code. But should be fine. Be aware that once you run it - because of the license of the server - it might be that people that use your service request the code and you need to provide it then to them. See AGPL license text for more details.

MorrisJobke:
And please use a lower case appid. (in the naming of the folder and file as well as in the CSR)

leonvandebroek:
@MorrisJobke I recreated the .csr using a lower case appid, I also just made changes to my app to conform with your suggestion to use lower case appid everywhere.

MorrisJobke:
55804ba

MorrisJobke:
@leonvandebroek There was a problem on our side when the certificates were generated. There are a few certificates that have the same serial number. For now this is nothing to worry about, but once one certificate gets revoked all certificates with the same serial number won't work anymore. Thus I would ask you to take this new certificate for your app: 7388ab1. You key is still the same and the old certificate still works, but it might be that due to a revocation in the future the current one might be revoked as well. So please use the new one. Sorry for the inconvenience.

PR #298, opened 2020-04-27
BigBlueButton integration

sualko:
[BigBlueButton](https://bigbluebutton.org) is a web conferencing system designed for online learning. My application allows to start sessions directly from the file browser with a presentation file.

https://github.com/sualko/cloud_bbb

cc @MorrisJobke


sualko:
It occurs that we can't add a csr via web, because Github doesn't sign off commits. Or did I miss something?

MorrisJobke:
> It occurs that we can't add a csr via web, because Github doesn't sign off commits. Or did I miss something?

sign-offs are not needed here, but we just set it up org wide to not need to configure it on every new repo. Unfortunately one cannot disable org wide checks per repo. So this is in here.

You can still sign-off, by simply place "Signed-off-by: NAME <EMAIL>" in the commit message. That's the whole technical part behind a sign-off :)

MorrisJobke:
a48efb5

sualko:
Thanks for the clarification and the fast merge :clap:

MorrisJobke:
@sualko There was a problem on our side. I would ask you to take this new certificate: be2bee2 The old one still works, but it might be that due to a revocation in the future the current one might be revoked as well. So please use the new one. Sorry for the inconvenience.

sualko:
No problem

sualko:
@MorrisJobke I updated the certificate, but I get "certificate revoked" now.

sualko:
Sorry, my fault.

MorrisJobke:
> Sorry, my fault.

No problem at all :) Good that it worked now :)

PR #309, opened 2020-05-25
Add richdocumentscode.csr

timar:


rullzer:
See https://github.com/nextcloud/app-certificate-requests/blob/master/richdocumentscode/richdocumentscode.crt

PR #315, opened 2020-06-11
Create quest.csr

antitoine:
@MorrisJobke Add sign request for publishing the Nextcloud Quest app.
You can found source of the Nextcloud Quest App here : https://github.com/antitoine/Quest-nextcloud

The goal is to establish access to our Quest (https://github.com/hverlin/Quest) application to search in a Nextcloud server for files. What was currently present as an open API in Nextcloud (Webdav) does not allow paging or links to the web space.


blizzz:
Here you go: 7011d30f755225908859ad56e8c40f851ae0561f

PR #318, opened 2020-06-24
Create cpanelmailsync.csr

ethernet-zero:
Hi @MorrisJobke,

This is a very simple app that I created out of necessity. It listens for user account events, and when a user is created or deleted or a password is changed, it creates, deletes or changes the password of the corresponding mail account in a cPanel instance. The app is a perfect match for the Mail app for administrators that use cPanel to provide mail accounts to their users and want to keep them in Nextcloud for simplicity.

The source code is here: https://gitlab.com/ethernet.zero/nextcloud-cpanelmailsync


MorrisJobke:
5075e5b

PR #320, opened 2020-07-27
Create loganalyzer.csr

lukaszhandy:
@MorrisJobke
https://github.com/lukaszhandy/loganalyzer


MorrisJobke:
@lukaszhandy Do you mind to fix some SQL injections first?

https://github.com/lukaszhandy/loganalyzer/blob/master/lib/Controller/PageController.php#L56
https://github.com/lukaszhandy/loganalyzer/blob/master/lib/Controller/DBController.php#L83

PR #330, opened 2020-08-24
Request cert for registration

nickvergessen:
As per https://github.com/nextcloud/registration/issues/210 pellaeon seems very busy at the moment.
So let's create a new cert and I will offer to give it to him as well when he is back again.


jancborchardt (approved):

MorrisJobke:
@rullzer We then need to withdraw the old one.

rullzer:
I did. that is why the CRL got updated

MorrisJobke:
> I did. that is why the CRL got updated

Thx

PR #331, opened 2020-08-24
Request cert for talk_matterbridge

nickvergessen:
Signed-off-by: Joas Schilling <coding@schilljs.com>


jancborchardt:
Just like in the GitHub, GitLab, Mastodon, etc requests – doesn’t it make sense to also call this `integration_matterbridge` since we could do stuff beyond Talk in the future?

If you disagree, feel free to just ignore the remark. ;)

nickvergessen:
No, this is just a way to easily manage and update the matterbridge binary across multiple branches and keep talks package size sensible (the binaries are 120MB). It does not come with anything else: https://github.com/nextcloud/talk_matterbridge

PR #340, opened 2020-09-01
add richdocumentscode_arm64.csr

timar:
@MorrisJobke please add this, thanks.
This is the ARM64 port of richdocumentscode.


juliushaertl:
@rullzer

PR #346, opened 2020-09-21
add integration_moodle.csr

eneiluj:
Moodle integration. Dashboard widget and search provider.

https://github.com/nextcloud/integration_moodle


eneiluj:
It would be nice to have this app in the appstore before Saturday. No hurry though :grin:

PR #347, opened 2020-09-29
Create solid.csr

michielbdejong:
CC @MorrisJobke

Intended app source code is here: https://github.com/pdsinterop/solid-nextcloud


rullzer:
Nice stuff
HEre you go: https://github.com/nextcloud/app-certificate-requests/blob/master/solid/solid.crt

PR #361, opened 2021-01-08
Create csp_editor.csr

tcitworld:
Add certificate request for CSP Editor

Code (soon) at https://framagit.org/framasoft/nextcloud/csp-editor

@MorrisJobke


tcitworld:
Forgot about DCO... Is the GH action really necessary here though?

nickvergessen:
Sorry, but why would you want to edit the CSP that easily? You really shouldn't. CSP rules are there for a reason.... Security reason. If you need to modify them you might be doing something wrong, or the app you use should properly expand the policy on it's own controllers. But allowing admins and hackers without a clue to modify the csp that easily in the frontend is a really bad idea.

tcitworld:
So these are the reasons that I can think of to edit the CSP Policy:
* we have custom home pages on https://framagenda.org and https://framadrive.org (through theme) that require external resources. This was actually the main reason for this app, as I was getting tired of editing `lib/public/AppFramework/Http/ContentSecurityPolicy.php` after each upgrade
* [another app of mine](https://apps.nextcloud.com/apps/login_notes) allows to insert custom markdown content on the login page and [it wouldn't work](https://framagit.org/framasoft/nextcloud/login-notes/-/issues/5) for external domain pictures¹
* it can be useful for specifying `frame-ancestors` to select which domains can embed a share link or public calendar page
* it can be used to avoid editing CSP to get [Vue DevTools to work](https://github.com/nextcloud/nextcloud-vue#using-vue-devtools-in-firefox)²

As you said all of these are very advanced and dangerous settings, but I thing there can be some legitimate cases where it might be useful.

¹ I could try to detect markdown pictures directly in the app and make the CSP rules here though
² Though allowing inline scripts & eval could be restricted to `DEBUG` in `config.php` for instance

nickvergessen:
1. We could fix via allowing the trusted domains
2. You could parse urls out and then allow only them within that app, instead of offloading the work to each admin (if it's not solved by 1 already)
3. (See 1.) Or a config dedicated for the framing csp
4. That doesn't warrant an app in the app store.

tcitworld:
> We could fix via allowing the trusted domains

This sounds really worse security-wise, as `trusted_domains` is used to get the server's host? These domains are not alternative names for my NC instance. In my case I control the domains, but what about people wanting to use 3rd-party domain for CDNs and stuff?
Also this would use `default-src` for the domain, but it would be more appropriate to have only `img-src` for Wikimedia Commons, or `font-src` for Google Fonts for instance, so that these domains can't do anything shady.

Anyway, I don't mind much not publishing it if sounds too powerful and dangerous.

*/insert SPIDERMAN_GREAT_POWER_GREAT_RESPONSIBILITY_MEME.jpg*

nickvergessen:
The problem i see is a hacked admin account allows weakening the security via web UI further. Let's see what Morris and Roeland think

tcitworld:
Ideas from the top of my head to improve security:
* Allow to change the settings within a grace period triggered by an OCC command (however, OCCWeb still exists)
* Allow to change the settings only if a setting is set inside config.php (I know, apps shouldn't touch this)
* Send emails to admins when settings are changed (however, emails can be disabled first by a hacked admin user)

MorrisJobke (approved):

MorrisJobke:
Obviously right after the merge the comments showed up, because I had the tab open 🙈

MorrisJobke:
> The problem i see is a hacked admin account allows weakening the security via web UI further. Let's see what Morris and Roeland think

Generally spoken hackers can use any other app and manipulate it in that way, right? Sure it's not a that general purpose approach. There only needs to be an app out there that adds a wildcard CSP and then you are screwed as well, or am I wrong? We can't forbid that other apps do those same weird security workarounds as well. And just blocking this one here, because it has it in the name itself, feels a bit like "security through obscurity".

On the other side it would make it harder and that is also a good thing.

tcitworld:
Waiting for @rullzer input to know if I can go ahead on this or not.

rullzer:
yeah each app can do it already...
Maybe just add a big warning sign?

tcitworld:
@MorrisJobke There's something wrong with the certificate, seems like you copied `csp_editor.csr` to `csp_editor.crt`.

MorrisJobke:
> @MorrisJobke There's something wrong with the certificate, seems like you copied `csp_editor.csr` to `csp_editor.crt`.

Yes 🙈 Let me push the crt.

MorrisJobke:
5c88f97e5e3495579fbebafe531ffe4f7e0d413f

PR #364, opened 2021-01-20
Renew all integration certs for repos under nextcloud namespace

eneiluj:
Just in case.


rullzer:
@eneiluj remember to rerelease all the apps with the new cert. Else people can't install them anymore.

PR #368, opened 2021-02-02
Add welcome.csr

eneiluj:
This app registers a welcome dashboard widget which renders an admin chosen markdown file and shows a link to call a support contact in Talk.

https://github.com/eneiluj/welcome

The app ID could be more explicit like "welcome_widget" or "welcome_dashboard" or even "welcome_message". I don't know yet if it will stay a widget-only app or if it will bring more stuff somewhere else. Better leave the ID independent from the dashboard. What do you think?


ChristophWurst:
> [eneiluj](/eneiluj) requested review from [rullzer](/rullzer), [MorrisJobke](/MorrisJobke), [ChristophWurst](/ChristophWurst), [wiswedel](/wiswedel) and [juliushaertl](/juliushaertl) [6 days ago](#event-4279013319)

I can't do this

PR #371, opened 2021-02-15
csr for "ehr" app

ayselafsar:
Signed-off-by: Aysel Afsar <aysel.afsar@gmail.com>


rullzer:
See https://github.com/nextcloud/app-certificate-requests/blob/master/ehr/ehr.crt

PR #375, opened 2021-03-02
Create user_unix_script.csr

bcecchinato:
Repo is not public yet, I need to clean it, but I would like to proceed to a first version soon. Thanks in advance @MorrisJobke or @rullzer.


rullzer:
See https://github.com/nextcloud/app-certificate-requests/blob/master/user_unix_script/user_unix_script.crt

PR #376, opened 2021-03-03
Add libresign

vitormattos:
App for digitally signing documents
https://github.com/LibreSign/libresign


rullzer:
See https://github.com/nextcloud/app-certificate-requests/blob/master/libresign/libresign.crt

PR #390, opened 2021-05-10
Hancom office connector csr file

TYRakeshN:
@MorrisJobke


MorrisJobke (changes requested):
The app id needs to be lower case. Do you mind to submit this again with a lower case app ID?

See https://github.com/nextcloud/server/blob/06377b7df1019f9287afd46f1202fdc86c0700f3/resources/app-info.xsd#L301-L305

TYRakeshN:
ill apply the changes & re submit

PR #411, opened 2021-07-03
Add Listman

revpriest:
Listman is a maillist manager https://github.com/revpriest/listman


blizzz:
Here you go 584a57b70cc2a2f6eee871dc81a69e2a410d9766

PR #412, opened 2021-07-05
CAS identity provider app

mziech:
This is an implementation of a CAS identity provider in order to seemlessly hook up 3rd party apps to Nextcloud authentication using the CAS protocol. It is not to be confused with the `user_cas` app which acting as a CAS service provider.

https://github.com/mziech/nextcloud-cas

Signed-off-by: Marco Ziech <marco@ziech.net>


blizzz:
here you go 1c822e038e231b3ca8dbb3c9525bef74837a9340

PR #421, opened 2021-08-20
Create talked.csr

mwalbeck:
Talked, call recording for Nextcloud Talk. This is a two part system, a Nextcloud app that acts as the client to control a server component that handles the actual recording.

You can find the source code for the Nextcloud app here: https://github.com/MetaProvide/talked-client

And the server source code here: https://github.com/MetaProvide/talked


mgallien:
you can find the certificate here 7ef4eff45977d2bf2044cf5eb13c662c5bde850d

PR #514, opened 2022-04-03
Create transfer.csr

danth:
App source code: https://github.com/danth/transfer


vitormattos:
Hi @danth !
I looked your repository, is a excellent idea!
Before follow the necessary to create your crt, I have a suggestion: maintain the name of repository equal to app_id will make more easy to new developers contribute to your app because only will be necessary clone the repository and start to work. In this case, rename from `nextcloud-transfer` only to `transfer` What you think about this?

danth:
@vitormattos Good idea, I've renamed the repository :+1:

PR #516, opened 2022-05-02
Create scimserviceprovider certificat

hrenard:
This app will be published by contact@indiehosters.net. The sources are [here](https://lab.libreho.st/libre.sh/scim/nextcloud-scim).
It's part of a [hackathon project](https://eventornado.com/submission/automatic-sso-saml-sync-from-identity-provider-keycloak-through-a-well-known-protocol-scim). So a quick review would be appreciated to publish at the end of the event.
[IndieHosters](https://indiehosters.net/) will be maintaining and improving this app. @pierreozoux is the lead on this.


vitormattos:
Good app! Congratulations!

About the review I don't know the best place to do this then, I will do here. If you like any topic, move the topic to project issues.

- Implement GitHub Actions
  - [ ] integration tests using Behat
    for example: https://github.com/juliushaertl/nextcloud-behat or https://github.com/LibreSign/nextcloud-behat
  - [ ] static analyse using psalm
      Do you will need require --dev `christophwurst/nextcloud` package to solve core issues and `vimeo/psalm`
  - [ ] linter using phpcs
      require --dev `nextcloud/coding-standard`
- GitHub Action to do the build and publish in app store
  I like this action to do this: https://github.com/R0Wi/nextcloud-appstore-push-action
- Put the repository with the name same as the APP_ID.
This will make more easy to a new developer start to code into the project because don't will need move the cloned directory to other place or clone the repository to other place. New developers only will need do a git clone inside the app folder.
- Clean the Controller folder
    The controller folder have two files that sounds like a Service file.
Will be good extend the controllers directly of ApiController or Controller class and move all logic from ASCIMGroup to `lib/Service/ASCIMGroup` by example
    https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html?highlight=service#making-things-reusable-and-decoupling-controllers-from-the-database
- Fix the file headers.
I think that all the file headers with the license was copied from other app. You can put with the real authors, assign the copright to main author and describe the list of co-authors.
- Put all routes in lower case.
    Is a good practice maintain all the routes in lower case. This will impact the integration.

> PS: I only will wait for you to get your opinion about  "Put the repository with the name same as the APP_ID".

hrenard:
Thanks for the review ! Indeed, we have some cleanup to do.
For the repo name, you mean the name or the slug (uri) ? It's not a problem.
We're implementing an RFC, so the route will stay as specified.

vitormattos:
I talked about the name of repository.
Will be good maintain the name equal to the APP_ID because when a new developer clone the project and if the project have the same name as the APP_ID, will more easy and the developer don't will need move the folder to a folder with the same name of APP_ID.

hrenard:
Okay, it's done.

vitormattos:
Here: [12760c239923114ed800bed9f7f0f2132f61cc78](https://github.com/nextcloud/app-certificate-requests/commit/12760c239923114ed800bed9f7f0f2132f61cc78)

PR #520, opened 2022-06-03
Create integration_dilingual.csr

LukasReschke:
`integration_dilingual` is a Nextcloud application that integrates with the [Dilingual](https://www.dilingual.com) website translation service. Currently the application shows the review state inside Dilingual on the dashboard.


vitormattos:
@LukasReschke do you have the repository of app?

MariusBluem:
The code for the app discussed here is currently not available in any public repository which we could publicly link here. Since this is an optional step I don’t think that this should be a blocker:

https://github.com/nextcloud/app-certificate-requests/blob/93306236070ed2391c8a370be708e6fe05a90250/README.md?plain=1#L9

Please let us know if there are any further questions about this request :)

vitormattos:
Here: 43492fb84e9762de7d66f242617b34e38b3a4f64

PR #525, opened 2022-07-07
picker.csr

eneiluj:
The picker app helps external systems to integrate Nextcloud by providing a page to pick a file and create a public link. It has some specific endpoints and pages to fit with the Webex embedded app flow.

https://github.com/eneiluj/picker


vitormattos:
crt file: https://github.com/nextcloud/app-certificate-requests/commit/e956dad8231272747a102929635ee419ed100d5b

eneiluj:
Thank you!

PR #531, opened 2022-07-26
add integration_miro.csr

eneiluj:
Work in progress: Miro integration that will probably look like the Nuiteq and the Excalidraw one:

https://github.com/eneiluj/integration_miro


vitormattos:
Crt file here: 7c47344f8c9d956fbf36a766e5e423481c9cabdf
one more app. uhuuu! :partying_face:

PR #532, opened 2022-07-26
add integration_excalidraw.csr

eneiluj:
Excalidraw integration that looks like the Nuiteq one:

https://github.com/eneiluj/integration_excalidraw


vitormattos:
The crt file: ecb41776e3d929f1d81c795957354c6efd478c53

vitormattos:
This isn't a man is an app machine.

eneiluj:
:joy: I'm only starting to implement those apps and some are very small :grin:

Thank you for the blazing fast answers! :heart_eyes: You are a crt machine!

PR #536, opened 2022-08-26
related_resources

ArtificialOwl:


nickvergessen:
Should be added to sysadmin key chain?
If we deploy from the releases org the key is not even needed to be released, right?
Cc @blizzz

blizzz:
> Should be added to sysadmin key chain? If we deploy from the releases org the key is not even needed to be released, right? Cc @blizzz

you mean given to the maintainer? That would not need to be necessary indeed.

ArtificialOwl:
@vitormattos is anything missing ?

vitormattos:
Hi @ArtificialOwl,
I don't found in a first moment what was the app but, searching here in the Nextcloud organization, I found:
https://github.com/nextcloud/related_resources
Generating now the crt file.

vitormattos (approved):

vitormattos:
CRT file here: cff51405ffe479b1dd85698208b90f1bf3038b7e

PR #597, opened 2023-04-17
We want to update our certifiacte

csoc-kevink:
### Checklist

- [x] Link to app repository: https://github.com/csoc-de/shifts


nickvergessen:
The repo URL is 404

It's of course okay to renew your cert/key, but we have to revoke the previous one and in the meantime your app might not be installable.
Did it get leaked or anything?

csoc-kevink:
Hey,

sorry for the typo. the correct URL ist https://github.com/csoc-de/shifts

No, the private key did not get leaked. We refactored the app to be compatible with v25 again. The old private key is not accessible anymore so we requested a new one.
That's not a problem. We're currently reviewing the update and when that is finished we will publish the new version.

vitormattos:
Hi @csoc-kevink,
Could you verify this first?
https://github.com/nextcloud/app-certificate-requests/pull/597/checks?check_run_id=12800904770

vitormattos:
@csoc-kevink you didn't remove the `shifts/shifts.crt` file, this will be replaced when we generate your cert. Could you revert this change?

csoc-kevink:
@vitormattos I've reverted the commit and only added the csr

vitormattos (approved):

vitormattos:
Your new crt file: 5a7548705bb187efb5a69055855951a70b9ba134

csoc-kevink:
Hi @vitormattos,

I think something went wrong with the signed certificate. We tried uploading our release bug we keep getting the error "data too large for modulus". I check our CSR and the signed CRT file and the public keys are not the same (see screenshot: signed cert left, CSR in repo middle and local CSR right).
Checksum of our local public key:       05 7F 95 A0 4D 1F B7 7D FD 93 4A 04 F6 A7 F5 30 54 F3 88 8C
Checksum of CRT remote public key: 6C BA 0F D0 25 DB E6 22 53 7D 7D 84 74 D9 5E 94 BA B2 21 29
Can you check if the corrected CSR has been signed?

PR #607, opened 2023-06-27
sketch_picker csr

julien-nc:
https://github.com/julien-nc/sketch_picker

This app is a smart picker provider to draw a sketch with a simple editor (same one as in the Viewer and Photos), then get a link to the sketch that has a link preview.


mgallien:
@julien-nc you can find the certificate here f2bd59505400a47f4f8a66e78d515454e62ee993

julien-nc:
Thank you!

PR #625, opened 2023-08-10
Add assistant csr

julien-nc:
https://github.com/julien-nc/assistant

This app provides a user interface to the new text processing API and allows other apps to integrate.


mgallien:
You can find the certificate here 29a42f335511805b1f3d944d822c449ced5f7a5f

PR #637, opened 2023-09-28
Create firstapp.csr

brenyodani:
https://github.com/brenyodani/rs4nc

### Checklist

https://github.com/brenyodani/rs4nc


mgallien:
@brenyodani I am confused by the chosen name for the app and the fact that this is not matching the repository name.
`FirstApp` is way to generic and not descriptive enough.
I could not really understand what I would do with your app.
I highly recommend having a more self explanatory name for this (and maybe with less generic words in it).
Let me know if I can help

PR #642, opened 2023-10-03
Added upscaler_example.csr

andrey18106:
This is an example external app for [app_api](https://github.com/cloud-py-api/app_api).

### Checklist

- Link to repository: https://github.com/cloud-py-api/upscaler_demo


nickvergessen:
Rename to `image_upscaler`

No need to label it example, if it can do the functionality

DaphneMuller:
Is this naming convention discussion really that relevant? These are just ID's. We're not violating any trademarks with this ID and the ID's are not suggesting any different functionality than what the app is about. I feel this slows down things a bit and that this discussion has a low priority.

mgallien:
@andrey18106 You can find the certificate here 420b7f10439a0a4c2699ab2a1ae55285eed84749

mgallien:
I also agree that the name are not ideal given they are indeed visible to admins and may convey the idea that they are pure examples not to use on production instances

PR #643, opened 2023-10-03
Added talk_bot_ai_example.csr

andrey18106:
This is an example external app for [app_api](https://github.com/cloud-py-api/app_api).

### Checklist

- Link to repository: https://github.com/cloud-py-api/talk_bot_ai_example (will be moved from [nc_py_api examples](https://github.com/cloud-py-api/nc_py_api/tree/main/examples/as_app/talk_bot_ai))


bigcat88:
> I don't think we want to distribute that example via the appstore? Otherwise it should have a more reflective name, because admins will see the entry?

We have a desire to have two working examples in the AppStore, one related to Talk + AI model (when later we can do UI, it will be easy to add model selection) and the second with just AI and image restoration.

We plan to have both examples working and to support them in the future, and they can be used by other people as a skeletons for much more complex stuff.

As for the name: this is just the name of the repository; the Bot itself will be called differently in the Talk settings.
We don't want to use `talk_bot_ai` in AppStore and to reserve it, so we choose adding `_example` to repos names to show that this is our official examples/skeletons and their modified source code can be used by anyone(MIT license I hope).

DaphneMuller:
Yes, we'd like to distribute app ecosystem v2 type of apps through the app store. Also other talk bots that don't make use of the app ecosystem v2 could be in the app store through registering it as an integration, I think it makes total sense to have all such extra possibilities for extra functionality listed in the app store.
@AndyScherzinger 

I don't think the discussion about the naming is relevant, this is just an ID.

AndyScherzinger:
I am fine with the ID as well as the distribution via the store which also gets/got extended for external apps.

So 👍 from my side

nickvergessen:
> Yes, we'd like to distribute app ecosystem v2 type of apps through the app store. Also other talk bots that don't make use of the app ecosystem v2 could be in the app store through registering it as an integration, I think it makes total sense to have all such extra possibilities for extra functionality listed in the app store.

Yeah totally up and in for that 👍

> I don't think the discussion about the naming is relevant, this is just an ID.

I just recommend not to use demo/example, as it might confuse admins when they see "examples" being enabled on production instances in a system report, while it actually can be real functionality users might miss when it because disabled 🤷

DaphneMuller:
for now this is intended to be an example app for app developers

mgallien:
@andrey18106 you can find the certificate here a0f07efe552a8fb1740e41619091e30814fdfb0a

PR #652, opened 2023-10-20
create nextcloud-aio cert in order to block it for AIO

szaimen:
### Checklist

For https://github.com/nextcloud/all-in-one/tree/main/app


juliushaertl:
This is not a valid app id:

id

        required

        must contain only lowercase ASCII characters and underscore

        must match the first folder in the archive

        will be used to identify the app

https://docs.nextcloud.com/server/latest/developer_manual/app_development/info.html

Also I would not have nextcloud in the app id, maybe rather use either `aio` or `all_in_one`?

szaimen:
> [This](https://github.com/nextcloud/all-in-one/blob/main/app/appinfo/info.xml?rgh-link-date=2023-10-20T10%3A43%3A26Z#L4) is not a valid app id:
> 
> id
> 
> ```
>     required
> 
>     must contain only lowercase ASCII characters and underscore
> 
>     must match the first folder in the archive
> 
>     will be used to identify the app
> ```
> 
> [docs.nextcloud.com/server/latest/developer_manual/app_development/info.html](https://docs.nextcloud.com/server/latest/developer_manual/app_development/info.html)
> 
> Also I would not have nextcloud in the app id, maybe rather use either `aio` or `all_in_one`?

Actually I am not planning to publish it on the store. @nickvergessen simply suggested to block it for later. It is included only in aio like this: https://github.com/nextcloud/all-in-one/blob/499dccb12255304b3124c4f787be169b053dad2e/Containers/nextcloud/Dockerfile#L209

szaimen:
Just discussed with @nickvergessen that it still makes sense for the case that we allow `-` at some point.

mgallien:
@szaimen you can find it here 058aa0d471b05be15ec484c5db33428732c7fb5f

szaimen:
Yes, thanks! I dont need it but glad to have it now :)

PR #657, opened 2023-11-10
csr for integration_docusign added

nc-fkl:
### Checklist

- https://github.com/nc-fkl/integration_docusign


integration_docusign.csr
```diff
@@ -0,0 +1,26 @@
+-----BEGIN CERTIFICATE REQUEST-----
```
nickvergessen, line 1: Should be in a folder with the appid as name

nc-fkl: @nickvergessen its done thanks - pls nickvergessen ;-)

mgallien:
@nc-fkl you can find it here 40b36c9efd41ba635e671bf51de20a652f2f9c53

PR #659, opened 2023-11-22
Add integration_swp.csr

julien-nc:
https://github.com/nextcloud/integration_swp

This is the app grouping all the Phoenix features. It won't be published in the appstore but we need a cert to sign the release archives.


mgallien:
@julien-nc you can find it here 5b7d9425512c6ee9e47cb9107e6edbc054f55c08

julien-nc:
Thank you @mgallien!

PR #681, opened 2024-02-05
Update documentserver_community certificate

chadek:
As previous maintainer is not active anymore on this project anymore, I submit my csr to start maintaining this [app](https://github.com/nextcloud/documentserver_community)


chadek:
I'm new to app publishing, previously spoke on this matter with @AndyScherzinger on [this issue of documentserver_community](https://github.com/nextcloud/documentserver_community/issues/298) repo. Hopefully it will make it's way. I also needed the app to be transfer to me on the app store, I used github sso to log on the store, account name should be the same.

PR #683, opened 2024-02-08
Update journal.csr

moffer:
Due to uppercase CN before.

### Checklist

- Please include a link to your app repository: https://github.com/moffer/nextcloud-journal


moffer:
@mgallien I am sorry for the inconvenience, while registering a release I have realized that I did not only write the csr file in uppercase, but also the CN name. This can't be uppercase as the documentation mentions.

mgallien:
@moffer 
no worries
you are not the first one to have this issue
we should also check this on our side to avoid those issues and we are not there yet (at least with tooling to automate the check)
please find your new certificate here e9e0bc0da5d583c297b5ed5c87da3bf169ee2b0b

PR #685, opened 2024-02-21
Create integration_paperless.csr

MB-Finski:
https://github.com/nextcloud/integration_paperless

Thanks for the help!


mgallien:
@MB-Finski you can find it here
sorry for the delay

PR #692, opened 2024-04-11
Create integration_nepenthes.csr

AaronDewes:
Nepenthes is an in-development project management software with a focus on performance. We're planning on making the first stable release soon, but would like to finish the Nextcloud integration before the release.

It is forked from OpenProject, so the integration_nepenthes is very similar to/based on `integration_openproject`.

https://gitlab.com/nirvati/nepenthes/nextcloud-integration


mgallien:
@AaronDewes you can find it here 90e1c3751253aeef2821bed002950c9437533617

PR #699, opened 2024-06-26
Request new certificate for whiteboard

juliushaertl:
This is a new certificate for an existing app id, that the author @mamatt agreed via mail to hand over to us :)

App ownership transfer for the app store is filed separately


mgallien:
@juliushaertl you can find it here 4b2c36b168abe0d7be0fa70dee08d1a4bf5ad4f8
old certificate is revoked

PR #700, opened 2024-07-07
Add mindmap.crt

admorgan:
Fork of files_mindmap at https://github.com/admorgan/files_mindmap


mgallien:
@admorgan I am not really sure having two apps with the same name will work fine.
Please bear with the fact that I am not an expert of the appstore itself.
Could you maybe discuss with the author of the app that you have forked and find a way to not have two apps with the same name ?

admorgan:
I wholeheartedly agree. Unfortunately the author is missing in action and has not responded to requests for a while. I would prefer not to fork, but I also use this app every day and the version in the parent repo does not work with NC after 27. See https://github.com/ACTom/files_mindmap/issues/191 and https://github.com/ACTom/files_mindmap/issues/197.

Would it be better if I just renamed the whole app to something different? I thought it would be disingenous as I plan to keep the features of the app the same.I am willing to do what it takes to keep using this app.

mgallien:
@admorgan I am looking into this (what would be needed to get you to be able to publish new releases)
sorry for the delay
did you try reaching out directly (via github and/or emails) to the original author ?
I can see that there is not much activity into those github issues

admorgan:
@mgallien I have attempted to DM and e-mail ACTom, but have not received any response in many months. I have another option that may work better for everyone involved. I can just change the name of the application to mindmap2. ACTom's vs version doesn't work with 28 and above. I can only release binaries for 28+ and then there wouldn't be any overlap despite there being functionally two of the same app. My worry is that ACTom eventually comes back and start making releases again some day. While I can update the name of the application, I am not sure how much work it would be to ensure both apps wouldn't interfere with each other if they were both installed at the same time.

mgallien:
> @mgallien I have attempted to DM and e-mail ACTom, but have not received any response in many months. I have another option that may work better for everyone involved. I can just change the name of the application to mindmap2. ACTom's vs version doesn't work with 28 and above. I can only release binaries for 28+ and then there wouldn't be any overlap despite there being functionally two of the same app. 

so from my understanding, renaming is up to you as long as the app id are different, both app can be published on app store

> My worry is that ACTom eventually comes back and start making releases again some day. While I can update the name of the application, I am not sure how much work it would be to ensure both apps wouldn't interfere with each other if they were both installed at the same time.

I would suggest to open a github issue on the original repository writing your plan and asking @ACTom what are their plans
that way, if he works again on it, he cannot miss what you are doing

admorgan:
Previous maintainer has returned.

PR #713, opened 2024-08-22
Create sessiontoken.csr

vincib:
The sessiontoken app, developped by Octopuce for Nextcloud gives an API endpoint to ask for sessiontoken for any user, similar to impersonate, but via an API. It's used to migrate nextcloud accounts massively or to access nextcloud accounts from other apps

The sourcecode is available here 

https://octoforge.fr/octopuce/sessiontoken


mgallien:
@vincib we are not really sure it would be a good idea to have such an app published in the app store.

The purpose of the app is enabling mass impersonating of users by administrators if I understand it correctly. We are not so sure that has a general interest that would justify having it on the appstore.

What do you think ?
Do you really think it would be of general use ?
We already are not fan of the impersonate app to be honest. The scope of it is smaller but could already be abused.

vincib:
Hello Matthieu,

I think about that issue the other way around: having good tools around an application is a good sign that it will be able to serve their user well.

To me, a free software is useful when it has 2 main characteristics: 
- a good documentation
- a wide ecosystem of tools to handle that software.

Nextcloud, as of today, fit those 2 well! 
An example: in 2005, I preferred using mysql over postgresql for this very reason: the documentation was better, and we already had circular replication in MySQL, a lot of tools (phpmyadmin, mysqldump, percona backup, fine-tuning of a lot of internals etc.) that were missing in pgsql at the time. 

As of Nextcloud, I don't know how I would be able to serve my users in the Nextcloud without the Impersonate app: the users know I can act on their behalf, and they are glad I can do that, since it means I'll be able to help them. 

The sessiontoken app is used in two main cases as of today: 

- To interact with other apps fluently (which means, without asking the user to re-authenticate, a better model UX-wise) like what we did for the visioconference software of the ministry of education in France: their user can use their NC files from their BBB instance without having to authenticate a second time.

- To migrate users massively from one Nextcloud to another without the need for very skillfull system administrators or ditry hacks 

Also, the sessiontoken app creates token in NC the "standard way" so the user can SEE the token in their "security" tab. 

I'd just suggest that the end user of a NC instance should KNOW when the impersonate app is installed (maybe in the security tab too?) 

Also, final point: if an admin (having ssh access to the NC instance) want to abuse their users, they can always install the impersonate app, whether it is in the app store or not... hiding it from the appstore would not protect anyone from rogue admins... 
So, to me, this kind of app clearly have legitimate use and should clearly be in the public appstores, so that anyone can see what's possible for the admins.

mgallien (changes requested):
@vincib we took some time to think again at the situation.
We would like to suggest that you should feel free to reach to us to discuss improvements to your solution.
You could use our forum https://help.nextcloud.com/
You could reach out via Talk community room https://cloud.nextcloud.com/call/xs25tz5y

We are concerned by this application from a security point of view. Those concerns are not big enough that we will block this app certificate request. We would still ask you to add some disclaimer to your readme giving context that more secure options are also existing.
As I said, we are eager to discuss this in a more direct with you and to resolve any issue.
Let me also remind you that we are organizing a conference very soon that may be interesting for you.
Do you consider joining us ?
https://nextcloud.com/blog/nextcloud-conference-2024/

PR #736, opened 2024-11-03
Create open_api.csr

smarinier:
https://github.com/smarinier/nextcloud_open_api/

As Nextcloud introduced recently the use of openapi.json in application folders, this app is intended to provide an easy Swagger UI interface for all these. It may be used as a documentation, or a "try out in situation".

This avoid (sometimes) complex Postman authentification configurations. The API calls are made "as in", meaning you can't do more than the current user is allowed to do. (No impersonate)

Of course admins may use the "Limit to groups" if necessary.


smarinier:
retry with my own email

PR #737, opened 2024-11-03
Create open_api.csr

smarinier:
https://github.com/smarinier/nextcloud_open_api/

As Nextcloud introduced recently the use of openapi.json in application folders, this app is intended to provide an easy Swagger UI interface for all these. It may be used as a documentation, or a "try out in situation".

This avoid (sometimes) complex Postman authentification configurations. The API calls are made "as in", meaning you can't do more than the current user is allowed to do. (No impersonate)

Of course admins may use the "Limit to groups" if necessary.


nickvergessen:
The name is a bit misleading, but https://github.com/nextcloud/ocs_api_viewer is basically doing the exact same thing 👍
Want to join forces there?

smarinier:
> The name is a bit misleading, but https://github.com/nextcloud/ocs_api_viewer is basically doing the exact same thing 👍 Want to join forces there?

Hi Joas,

Ahh... Too bad I didn't find it by searching with the "Application store search". I've built it for my own use. And then I've spent half one day to make it "publishable in the store", considering it may help... as it wasn't in the store.

What i've (quickly) seen as main differences, is that:
 - my sidebar is a bit more pleasant (my subjective opinion ;) ) with app icons and names
 - I bypass any authentication troubles, as it's basically the current nextcloud user that makes the request (may be an inconvenient or an advantage, depending on the point of view). (In fact, i don't know it the basic auth is used in ocs_api_viewer or if it's kept because of a lack of its api).
 - I don't use iframe as the swagger UI component works fine in the NC context
 
 Inconvenient:
  - I'm not as much exhaustive, as I only look at openapi.json (but there seems to be some redundancy with -full, or other json).
 
 Hi @provokateurin (we met in Berlin on month ago). Let me know what you think about it.

provokateurin:
Thanks for building this app @smarinier!
I think we should merge the efforts.

> Too bad I didn't find it by searching with the "Application store search".

The store search is not great and for marketing reasons it had to be named "OCS" and not "OpenAPI" (which I still dislike but here we are anyway)

> my sidebar is a bit more pleasant (my subjective opinion ;) ) with app icons and names

This can be implemented in the ocs_api_viewer as well.

> I bypass any authentication troubles

This is great and it would also solve https://github.com/nextcloud/ocs_api_viewer/issues/3

> I don't use iframe as the swagger UI component works fine in the NC context

Maybe this can also solve https://github.com/nextcloud/ocs_api_viewer/issues/34?

> I'm not as much exhaustive, as I only look at openapi.json (but there seems to be some redundancy with -full, or other json).

Yeah this should be checked as well for completeness.

Do you think it's feasible for you to contribute these things to the ocs_api_viewer? I'm happy to make the switch to the swagger UI, if it can bring these benefits.

smarinier:
Bi @provokateurin, I'm fine with that, but, could you confirm:
Do you agree on avoiding the identification capability, as I've done in my app ? The principle is "you can do with the API (eg ocs_api_viewer) only what the current authentified session user is allowed to do". Is that ok with you ?
To my opinion:
 -  you can anyway, through the Swagger UI interface, get the openapi JSON content (wich you couldn't get before except by accessing to the code files tree), you import it into Postman, if you want to do more complex API stuff.
  - Basic Auth is not always the authentication solution used on the NC platform
  - Having a session on the same domain with another authentication trough an iframe may produce some strange things, and probably a potential security breach ?

provokateurin:
>  Do you agree on avoiding the identification capability, as I've done in my app ? The principle is "you can do with the API (eg ocs_api_viewer) only what the current authentified session user is allowed to do". Is that ok with you ?

Yes, I think that is better anyway.
The app is meant for developers which are going to have admin access anyway on their dev instance, so it doesn't really matter.Yes, I think that is better anyway.

PR #693, opened 2024-04-24
groupfolder_filesystem_snapshots certificate request

JonathanTreffler:
- App Name: Groupfolder Filesystem Snapshots
- Owner: @verdigado (you can check my affiliation with my public membership in the github org and github verified verdigado email address)
- Git Repository: https://git.verdigado.com/verdigado-public/nextcloud_groupfolder_filesystem_snapshots
- License: agpl

PR #739, opened 2024-11-18
organization_folders certificate request

JonathanTreffler:
- App Name: Organization Folders
- Owner: @verdigado (you can check my affiliation with my public membership in the github org)
- Git Repository: https://github.com/verdigado/organization_folders
- License: agpl

This app I am very excited for: Ever noticed groupfolder ACLs are impossibly complicated to manage by hand at a very large scale (thousands of groupfolders) and ACL editing can't be delegated in a granular enough way ?

This app is a complete layer of abstraction over groupfolders with an end-user friendly UI and a permissions system, that is a lot simpler and easier to wrap your head around than raw ACLs, that get translated into regular groupfolder ACLs in the backend.
But that is not all: Other apps can register themselves as organization providers, that provide a standardized, structured model of an organization, so organization folder managers can authorize (Sub)Organization Members and users with organization roles (both backed by groups) instead of a having to authorize individual users or picking an group from a unstructured list of groups (which, for the kind of instance this was made for would be thousands).

This is a mostly-rewritten (backend completely, frontend not completely) evolution of an "internal" app, that was written for one organization specifically. It was always AGPL3, but it was completely useless for anybody else, because it was so specific to that organization. But now it is general-purpose, ready to be used by other organizations: You only need to write an organization provider to adapt it or disregard the whole organization system and use it with users and groups only.

It will still be a few weeks of development, before I want to release a beta of this app to the appstore (maybe nightly releases before that), but I wanted to get this certificate request out of the way well before that.
Also this app depends on https://github.com/verdigado/groupfolder_tags, for which I will create a certificate request soon (Edit: see #740).

P.S.: I guess it now makes more sense what https://github.com/nextcloud/app-certificate-requests/pull/693 is for :) This has been quite a while in the making.


GretaD:
hi, heres your certificate 47442c664514a91fd9de00438042f9041f96da6a

JonathanTreffler:
Thanks :)

PR #753, opened 2025-01-08
add the certificate for my webtransfer app

AlexandreBRAVO:
Link to the repository : 

https://github.com/Pl4yep2on/webtransfer


mgallien (changes requested):
@AlexandreBRAVO sorry to bother you but could you tell me what this app is intended to do ?
I understand that we usually do not ask for this but from the current state of your app, I failed to understand what it should do

AlexandreBRAVO:
Good evening @mgallien , sorry i didn't update the readme on the main branch of the app. So to keep simple, the app i developped basically add a new route to a nextcloud server. This new route needs as parameter an url to a file (any type) and then on the page which the route leads to, there are 2 parts :
 - on the left : the file you sent the corresponding url
 - on the right : teh files of the connected user on the nextcloud server

Then the user can drag and drop the file anywhere he wants in his personnal space.
And this is the use case for any type of file. But what we developp the app for is for zip files specifically. If you send an url to a zip file, when you get the page to drag and drop your file, on the left you don't have the zip file, but its content like if it was unziped. And then you can transfer the files you need in the zip archive.

I don't know if i made it more clean. Ask me if you want more informations about it or a better explanation.
Have a nice day.

mgallien (approved):

mgallien:
@AlexandreBRAVO thanks for the explanation
I am very sorry for the too long delay
please find the certificate here 22745b2250ee4731344b3c5a69c37fcd73dd3d4a

PR #757, opened 2025-01-16
Create user_flow.csr

bvid34tgd:
### Checklist

- Please include a link to your app repository https://github.com/bvid34tgd/user_flow


mgallien (changes requested):
@bvid34tgd I do not really understand what is the purpose of your app especially when compared to https://github.com/nextcloud/impersonate
as the app is all about impersonating normal users, I think the name should better reflect that
I would as an administrator be very confused as to what is the selling point of your app especially with regard to alternatives
please keep also in mind that impersonating users is a very sensitive topic with regard to security
this security aspect is why I would prefer to better understand the context of this PR before generating the certificate

PR #758, opened 2025-01-18
Create c.kiminox.net.csr

kiminox:
### Checklist


mgallien (changes requested):
@kiminox please include a link to your app's repository ?

mgallien:
@kiminox we think that using your domain name as the id for the app certificate is wrong
see https://docs.nextcloud.com/server/latest/developer_manual/app_development/info.html
the id of the app should follow this rule:
> must contain only lowercase ASCII characters and underscore

PR #745, opened 2025-01-01
Add CSR for photoframe app

jeppester:
**Link to app:**
https://github.com/jeppester/nextcloud-photo-frame

**Description (also in README):**
Photo Frame lets you generate photo frame URL's for your albums which can then be viewed on any screen with a browser.

The possiblities are endless:
- Use kiosk browsers (like Fully Kiosk Browser) to turn any android device into a photo frame
- Use collaborative photo albums to co-create and share photo frame content across households

**Important**
The app is functional in the way that I'm using it in my own instance, but it's not 100% ready to be published yet (for instance it lacks a bit of styling).

I wanted to start the process of publication process before spending more time on polishing the app - in case you don't want it in your store or there are other issues that need solving.

I have a few of questions that I'll humbly ask as part of this PR:
1. How can I make my app dependant on the photos app? (Is there a clever way to not allow installing the app unless photos is available?)
2. Currently I generate my own links by prefixing "/index.php/apps/photoframe/" to each link. Should I be doing something different?
3. Photo Frame generates a public link for each photo frame (a non-guessable link with a 64 char alphanumeric token). The URL only lets visitors fetch the current photo, so it's only possible to fetch 1-24 photos per day (depending on the photo frame's setting). Nextcloud has a global settings page for sharing. Should my photo frame app adhere to those settings? And if so, which setting and how do I get the current setting?
4. Must I add a [database.xml](https://nextcloudappstore.readthedocs.io/en/latest/developer.html#database-xml)?

PR #774, opened 2025-03-30
Add CSR for photo_frames app

jeppester:
**IMPORTANT:**
This is the same app as https://github.com/nextcloud/app-certificate-requests/pull/745, I changed the ID and figured that I'd need to create a new CSR. Otherwise I believe the app is ready for its first release :rocket: 

Should I create a PR for removing the old CSR or can you delete the CSR and certificate for me?

**Link to app:**
https://github.com/jeppester/nextcloud-photo-frames

**Description (also in README):**
Generate easily sharable photo frame URLs for your nextcloud albums.

The possiblities are endless:

- Turn any android device into a smart photo frame (with the help of a kiosk browser)
- Co-create photo frames with nextcloud's collaborative albums
- Share photo frames across households


mgallien (changes requested):
@jeppester yes
the best would be to add a commit in this PR that removes the older certificate and CSR files

jeppester:
> @jeppester yes the best would be to add a commit in this PR that removes the older certificate and CSR files

Done.

I apologize for causing you extra work by changing my mind regarding the name.

I figured it was better to get it right the first time than to live with the inconvenience of "legacy" right from the get-go.

jeppester:
> @jeppester sorry for the contradictory demands I thought I could use my internal tool to revoke the older certificate and generate the new one in fact this is not possible so can you split this into 2 PRs one would remove the older csr file one add the new csr file 🙏

No problem, and done.

The other PR is here:
https://github.com/nextcloud/app-certificate-requests/pull/780

mgallien:
@jeppester you can find the new certificate here 9df36944155628e93c6c9707ca1750f31493ed96

PR #777, opened 2025-04-02
Certificate request for integration_oidc

mickenordin:
Source code here:
* https://github.com/SUNET/nextcloud-integration_oidc/


mgallien (changes requested):
@mickenordin can you fix the failing DCO check ?

vitormattos:
There are already some apps for authentication with OIDC:

https://apps.nextcloud.com/?search=oidc

Are existing apps for OIDC not doing what you need?

I ask because it's always interesting to add efforts to an existing app rather than having several apps that do the same thing.

And, if the existing ones don't have a feature they want, wouldn't it be interesting to propose the feature to one of them and only then propose a new app if there is no success and the particularities of the new version make sense.

mickenordin:
This app is not about logging in to Nextcloud with oidc, or using Nextcloud as an oidc provider, but about integrating against oidc providers and managing accesstokens in a generic way so you can build other oidc aware apps without having to implement the oidc flow yourself.

Den 3 april 2025 01:33:19 CEST, Vitor Mattos ***@***.***> skrev:

mickenordin:
> @mickenordin can you fix the failing DCO check ?

@mgallien fixed

mgallien:
@mickenordin you can find the certificate here 2933f8f0efa5ca13266e2c559a0654f31d7b2be0

PR #779, opened 2025-04-03
Create lectorapp.csr

pp2s:
### Checklist


mgallien (changes requested):
can you close one of your 2 PRs ?
#779 or  #781 
maybe you want to keep #781 that seems to be having the single commit with the proper file in it

PR #780, opened 2025-04-03
Remove photoframe CSR and cert

jeppester:
As requested here:
https://github.com/nextcloud/app-certificate-requests/pull/774#issuecomment-2772484143


jeppester:
> @jeppester I already revoked the old certificate but can you fix the failing DCO check such that I can merge this PR that would be cleaner

I believe it's ready now

PR #796, opened 2025-05-22
Update whmcsintegration.csr

whmcs-global:
I want to register same application again because last application files were deleted from our website as we migrated to new server. So, our customers requesting they want to download app from nextcloud app store and getting issue so want to register new application with same name

That's why submitted PR with new certificate

### Checklist

https://github.com/whmcs-global/whmcsintegration


mgallien:
@whmcs-global you can find the new certificate here 2dcdd217d246403b9a629a1f0aa952301beb6555

PR #815, opened 2025-08-21
Create helloworld.csr

MajoMarian:
### Checklist

https://github.com/MajoMarian/helloworld


mgallien:
@MajoMarian I am curious what you want to achieve
do you really intend to publish it as a complete application or something else ?

MajoMarian:
Hello @mgallien this is just for the test purposes. We are running locally nextcloud Appstore and we would like to test it if we are able to register an app. If I understood correctly without certificate, it is not possible to register the app.

PR #833, opened 2025-09-30
feature/check-csr-subject

wilkis3:
This feature request will check the subject from the csr files in pull request and if the subject is equal to the filename

### DIFF
--- a/.github/workflows/merge_request_app_id.yml
+++ b/.github/workflows/merge_request_app_id.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
-        uses: actions/checkout@v4
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 #v5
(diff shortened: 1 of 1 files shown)


.github/workflows/merge_request_app_id.yml
```diff
@@ -0,0 +1,32 @@
...
+  pull_request:
+    paths:
+      - '**/*.csr'
+
+jobs:
+  validate-csr-cn:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
```
mgallien, line 12: can you replace it with the hash ?
otherwise that is a security issue as it happened that tag got compromised

mgallien (changes requested):

.github/workflows/merge_request_app_id.yml
wilkis3: @mgallien thank you for your tip. I added the hash and update to action version to v5

PR #835, opened 2025-09-30
Add audiocovcerpreview CSR

david-brauer:
The app is very basic but works very well for my music collection on my server
https://github.com/david-brauer/audiocoverpreview/blob/main/appinfo/info.xml


david-brauer:
You are super fast.
Thank you

PR #871, opened 2025-12-26
Add naming convention check

callmemagnus:
Hello,

Not your usually PR.

You shouldn't have to handle PR where naming convention are not followed!

I'm not sure about the convention. I set lowercase letters, numbers and underscore.

:-)

### DIFF
--- a/.github/workflows/validate_folder_names.yml
+++ b/.github/workflows/validate_folder_names.yml
@@ -1,0 +1,65 @@
+name: Validate Folder Names
+
+on:
+  pull_request:
+
+jobs:
+  validate-folders:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 #v5
+
+      - name: Validate folder names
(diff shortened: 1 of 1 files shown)


mgallien (changes requested):
@callmemagnus thanks
sorry for the delay
could you do a small change and we eventually merge this
I have one request where the check would have been useful

callmemagnus:
>  I have one request where the check would have been useful

What would you like to change ?

.github/workflows/validate_folder_names.yml
```diff
@@ -0,0 +1,68 @@
+name: Validate Folder Names
+
+on:
+  push:
+    branches:
+      - '**'
```
mgallien: I think we only want to run this on pull requests

mgallien (changes requested):
so apparently the comment was added but not recorded
sorry for the confusion
see my inline comment

.github/workflows/validate_folder_names.yml
callmemagnus: OK, done.

mgallien (approved):

PR #974, opened 2026-04-06
Add certificate request for odf-kit-service

GitHubNewbie0:
ExApp: odf-kit Service — exports Markdown, HTML, and plain text files to ODT format. Repo: https://github.com/GitHubNewbie0/odf-kit-service


mgallien (changes requested):
@GitHubNewbie0 the app ID is invalid
there are strict naming rules for app IDs
see https://nextcloudappstore.readthedocs.io/en/latest/developer.html#info-xml
you should use `_` instead of `-`

mgallien (changes requested):
@GitHubNewbie0 can you remove the first csr file (i.e. `odf-kit-service/odf-kit-service.csr`) ?
ideally you would have only one commit with the file with the correct name

mgallien (approved):

GitHubNewbie0:
Thank you Matthieu for your kind and patient assistance with this process!  I appreciate.

PR #1026, opened 2026-06-03
Add certificate request for doriath

rjzondervan:
https://codeberg.org/Conduction/doriath


mgallien (changes requested):
The app id is wrong and the default one from Nextcloud app template.
[Especially, `-` character is forbidden.](https://codeberg.org/Conduction/doriath/src/branch/main/appinfo/info.xml#L4)
See documentation on the rules here https://nextcloudappstore.readthedocs.io/en/latest/developer.html#obtaining-a-certificate

rjzondervan:
@mgallien That should be fixed now. Filling out the AppInfo had not yet proceeded to main.
Apologies for the noise by the way, we are under some pressure to get our apps in the store before tuesday, and that unfortunately results in you finding out, rather than us, what repositories still have issues.

mgallien (approved):

mgallien:
@rjzondervan Your crt file: 561e18b9decab683280a1c38b5b2d57d190915f7

PR #1048, opened 2026-06-17
Create desktop.csr

canisdata:
certficate request for app named desktop

https://github.com/canisdata/desktop


mgallien (changes requested):
@canisdata 
could you rename the app ID to something like `desktop_workspace` ?

We think that the use of `desktop` only may create some confusion with the desktop sync and talk clients for Nextcloud and would prefer to avoid that

canisdata:
the url has also changed to:
https://github.com/canisdata/desktop_workspace

mgallien:
@canisdata you can find the new certificate here 6217fac558db14a58b0d836751eca2038f98f93a
thanks for taking into account our request to rename

PR #1085, opened 2026-07-09
Add certificate signing request file

lindesbs:
### Checklist

https://github.com/lindesbs/app-certificate-requests


mgallien (changes requested):

PR #1090, opened 2026-07-13
Add certificate signing request for group photo sorter

fsch-ppi:
Code is at: https://github.com/fsch-ppi/group-photo-sorter/


mgallien (changes requested):
`<app ID>/<app ID>.csr`

can you fix it ?

you forgot to use the correct folder name 😄 

fsch-ppi:
@mgallien Thanks for your prompt reply, just fixed it as described!
Thanks so much!

mgallien (changes requested):
the issue comes from your second commit that is doing the rename
you might want (if you are comfortable with the tool) to squash all commits into a single one with correct DCO

PR #1115, opened 2026-07-28
Add csr for atrium_secureshare

jbrunner:
Share files with people outside your organisation directly from Nextcloud, without giving them Nextcloud accounts and without needing to expose Nextcloud to the public internet.

https://github.com/atrium-secureshare/atrium-plugin-nextcloud


mgallien:
@jbrunner you can find the certificate here 5156f5a3df3379b0742db2c7d63a03d201a67c8b

PR #1120, opened 2026-07-29
Add certificate request for sharepath

arundne:
Certificate request for the app id `sharepath`.

**Source code:** https://github.com/arundne/nextcloud-sharing-path

**Background:** This is a maintained continuation of the abandoned `sharingpath` app (last store release 0.4.4 in 2022, declares `>=20 <25`). It serves shared files by their stored path so they can be embedded directly. The store version has been broken since Nextcloud 25 and produces an *Internal Server Error* on every download since Nextcloud 33, because it still uses the private `OC_Response` class that was removed from the server.

The fixed version has been rebuilt on the public OCP API and tested against Nextcloud 33.0.6 (downloads including range and multipart range requests, the file action in the Vue files app, settings pages, 404/403 handling). It declares Nextcloud 25–34.

I have also offered the original author a PR and an ownership transfer of the existing entry ([rookie0#58](https://github.com/rookie0/nextcloud-sharing-path/pull/58), [rookie0#59](https://github.com/rookie0/nextcloud-sharing-path/issues/59)); there has been no response since 2022. This request is for a separate app id so users on managed hosting — who can only install from the App Store — have a working option in the meantime. The app ships its own PHP namespace so it can coexist with the old app on the same instance.


mgallien (changes requested):
@arundne 

arundne:
Thanks for the review @mgallien — the commit is amended with the `Signed-off-by` line and the DCO check passes now.

The CSR itself was generated exactly as documented: `openssl req -nodes -newkey rsa:4096 -keyout sharepath.key -out sharepath.csr -subj "/CN=sharepath"`. Let me know if anything else is needed.

arundne:
@mgallien friendly ping — this should be ready for another look.

- The DCO check passes.
- The two force-pushes after your review only changed the commit author's email address (and the matching `Signed-off-by` line); the CSR file itself is unchanged since the PR was opened.
- CSR generated exactly as documented: `openssl req -nodes -newkey rsa:4096 -keyout sharepath.key -out sharepath.csr -subj "/CN=sharepath"`

Could you re-review when you get a chance? Happy to adjust anything else that's needed. Thanks!

mgallien (changes requested):
The app id in the app repository is different from the request
That will prevent publishing it on the app store
see https://github.com/arundne/nextcloud-sharing-path/blob/master/appinfo/info.xml#L4
See documentation https://nextcloudappstore.readthedocs.io/en/latest/developer.html#obtaining-a-certificate

arundne:
Thanks @mgallien — good catch, and you were right.

The sources declared `sharingpath` and only produced the `sharepath` variant at build time, which of course isn't visible when you look at the repository. I've turned that around: [`appinfo/info.xml`](https://github.com/arundne/nextcloud-sharing-path/blob/master/appinfo/info.xml#L4) now declares `<id>sharepath</id>` — matching this request — and the build script generates the old `sharingpath` id as a variant instead, kept ready only in case the original App Store entry is ever transferred to us.

Background, for context: the original `sharingpath` app has been unmaintained since 2022 and returns an *Internal Server Error* on every download since Nextcloud 33 (it still uses the removed private `OC_Response` class). I offered the fixes and an ownership transfer to the original author ([rookie0#60](https://github.com/rookie0/nextcloud-sharing-path/pull/60), [rookie0#59](https://github.com/rookie0/nextcloud-sharing-path/issues/59)) — no response so far. Hence this request for a separate app id, so users on managed hosting have a working option.

Ready for another look whenever you have time.

nilsding (approved):

nilsding:
@arundne your new certificate can be found at a9e80f103f93a969787376edd358200328c5cbd7

arundne:
@nilsding could you take care of this, too?

PR #1124, opened 2026-07-30
Add certificate signing request file for Curio

Acknar:
[https://github.com/Acknar/Curio](https://github.com/Acknar/Curio)

An app to curate picture, videos, urls, pdfs in a board and map view for easy viewing and collaboration


mgallien:
@Acknar you can find the certificate here a5b5247ead365b766a0f9a7daf3c823702340830

PR #1136, opened 2026-08-04
Rotate opencatalogi certificate

WilcoLouwerse:
The previous opencatalogi certificate has been revoked by Nextcloud
(POST /api/v1/apps registration returns HTTP 400 "Certificate is invalid:
certificate revoked"). This PR replaces the CSR with a freshly generated
one so a new certificate can be issued.

Source: https://github.com/ConductionNL/opencatalogi


MWest2020:
@mgallien previously I asked for the certs, but the responsibilities are passed to other team members.

mgallien:
@WilcoLouwerse you can find the new certificate here 0192557a2329415b1b99c6a0ea8b8419ed5819b1

rubenvdlinde:
Follow-up on this rotation, in case it saves someone else the same trip.

The 403 that prompted it, "Only the app owner is allowed to update the certificate", was not a certificate problem. It is what the store returns when the API token used by CI belongs to an account that does not own the app. Our release pipeline was authenticating as an account that was not the owner of `opencatalogi`, so no new certificate could have fixed it. That has since been resolved on our side by publishing with the owning account, and `opencatalogi` released successfully again today.

As a result the certificate issued here on 17 August 2026 was never registered. apps.nextcloud.com still holds the certificate from 13 May 2026, which is the one our existing releases are signed with, and we are deliberately leaving it that way: registering a different certificate would delete every existing release of the app.

So this certificate is unused and will stay unused. Would you prefer to revoke it so it cannot be adopted by mistake later, or leave it dormant? Either is fine by us, we would just like the record to be accurate.

mgallien:
@rubenvdlinde I would assume that the older certificate is no longer valid to publish new releases
I thank you will need the new certificate to publish releases that are considered valid by server releases and the apps store

PR #1137, opened 2026-08-04
Rotate openregister certificate

WilcoLouwerse:
The openregister GitHub Actions release workflow returns HTTP 403
"Only the app owner is allowed to update the certificate" when trying
to register/update the cert on apps.nextcloud.com. Previous releases
were uploaded manually; we're moving to automated releases and need a
fresh certificate that matches the CI credentials.

This PR replaces the CSR with a freshly generated one so a new
certificate can be issued.

Source: https://github.com/ConductionNL/openregister


MWest2020:
@mgallien previously I asked for the certs, but the responsibilities are passed to other team members.

mgallien:
@WilcoLouwerse you can find the new certificate here 6a87812ebea1558f91c3c7e32f8e7c6056c67103
to transfer an app (and possibly changing the certificate), you have some documentation on the steps here https://nextcloudappstore.readthedocs.io/en/latest/developer.html#transferring-your-app-to-a-new-owner

rubenvdlinde:
Follow-up on this rotation, in case it saves someone else the same trip.

The 403 that prompted it, "Only the app owner is allowed to update the certificate", was not a certificate problem. It is what the store returns when the API token used by CI belongs to an account that does not own the app. Our release pipeline was authenticating as an account that was not the owner of `openregister`, so no new certificate could have fixed it. That has since been resolved on our side by publishing with the owning account, and `openregister` released successfully again today.

As a result the certificate issued here on 17 August 2026 was never registered. apps.nextcloud.com still holds the certificate from 13 May 2026, which is the one our existing releases are signed with, and we are deliberately leaving it that way: registering a different certificate would delete every existing release of the app.

So this certificate is unused and will stay unused. Would you prefer to revoke it so it cannot be adopted by mistake later, or leave it dormant? Either is fine by us, we would just like the record to be accurate.

mgallien:
@rubenvdlinde the old certificate should no longer be usable as it was revoked as part of generating the new one
did you manage to publish a release signed by the older certificate ?
the use of the latest (not revoked) certificate is only when publishing new releases older already existing ones still work

PR #1140, opened 2026-08-06
Revoke and replace certificate for sharepath (private key exposed)

arundne:
Requesting revocation and replacement of the certificate for the app id `sharepath` (issued in #1120).

**What happened:** the private key was accidentally shipped inside the published `0.7.1` release archives. My build script mirrored the whole working tree and filtered by exclusions; the directory holding the key was not on that list, and `.gitignore` only keeps a file out of git, not out of the release archive. The archives were public on the GitHub release page for about eight days and were reachable through the store entry.

**Already done on my side:**
- the affected archives are deleted from the release page (the release now has no assets)
- the key was never committed to the repository, and the earlier `0.7.0` archive predates it
- a fresh 4096 bit key pair was generated; this PR replaces `sharepath/sharepath.csr` and deletes `sharepath/sharepath.crt` as the documentation describes
- the build script now works from an explicit allowlist of what belongs in a Nextcloud app, so nothing outside it can end up in an archive again

No release signed with the old key was ever successfully installed anywhere — the store release failed on an unrelated packaging problem, which is also fixed in the meantime.

I'll re-register the app with the new certificate once it is signed, which drops the existing releases. Sorry for the extra work, and thanks.


nilsding:
@arundne the new cert can be found at 4d7da9a3fc6752e174cc8771b5a9746a22e3405b

PR #1146, opened 2026-08-09
Add certificate request to simplecontacts

dpfpic:
### Checklist

- https://github.com/dpfpic/simplecontacts


mgallien (changes requested):
`<app ID>/<app ID>.csr`

can you fix it ?

PR #1147, opened 2026-08-09
add cherificats to providers

dpfpic:
### Checklist

- https://github.com/dpfpic/providers


mgallien (changes requested):
`<app ID>/<app ID>.csr`

can you fix it ?

please use a less generic app ID ?
`providers` is definitely too generic

mgallien (changes requested):
There are now 2 certificate signing request files.
If I get it correctly you want to replace providers/providers.csr by simplecontacts/simplecontacts.csr
Can you just keep the correct file and maybe remove the extra commits for simplicity ?

mgallien (changes requested):
there are still two certificate signing request files in the PR including "providers.csr" which name is too generic
can you keep only one and use a less generic name ?

PR #1165, opened 2026-08-15
Add certificate request for findling

street1983nk:
findling is the PHP companion app of a Nextcloud search integration. It registers a search provider so that document search results appear in the normal unified search, and forwards every query to its ExApp counterpart findling_backend, which is submitted in a separate request. Source code: https://github.com/street1983nk/nextcloud-search, licensed AGPL-3.0-or-later.


mgallien:
@street1983nk you can find the certificate here 66055b90d743664050fc73fdf568826838ef0336

PR #1166, opened 2026-08-15
Add certificate request for findling_backend

street1983nk:
findling_backend is the ExApp container behind the findling search app. It extracts text, runs OCR and maintains the search index inside the instance, so no file content ever leaves the server. It is only reachable through the companion app findling, which is submitted in a separate request. Source code: https://github.com/street1983nk/nextcloud-search, licensed AGPL-3.0-or-later.


mgallien:
@street1983nk you can find the new certificate here 5da99da3b8af7f9bb22c79399afaf6a79c8ea72c

PR #1183, opened 2026-08-24
Add certificate signing request for netbase

ktec-jp:
Certificate signing request for the app **netbase** (NetBase — network toolbox for Nextcloud).

- App ID: `netbase`
- CN in the CSR matches the app ID
- Source: https://github.com/ktec-nc-apps/NetBase
- Author: KTEC (ktec-jp)

Only `netbase/netbase.csr` is added.


mgallien (changes requested):

ktec-jp:
@mgallien Thank you for the review, and sorry for the extra work. The DCO failure was entirely my fault: I created the commit through the GitHub API, so it carried no `Signed-off-by` line. I have amended the commit with a proper sign-off and force-pushed it, and the DCO check now passes. Sorry for the noise, and thanks for your time.

ktec-jp:
Friendly ping @mgallien — the DCO check is green again after the sign-off fix. The two validation workflows (`Validate Folder Names`, `Validate CSR Common Name`) are still waiting for a maintainer to approve the run, and the earlier change request is still standing, so the PR cannot move forward on my side. Could you take another look when you have a moment? Thank you.

ktec-jp:
Hi @mgallien — I see the queue is moving again today, thank you for working through it.

A small request for this one: the change you asked for is done. The DCO check has been green since 26 August (signed off locally with `git commit --amend -s`), and there is nothing else outstanding on my side. The PR is still blocked only because the `CHANGES_REQUESTED` review from 25 August is still standing.

Could you please dismiss or re-submit that review when you get a moment, so it can be merged? Thanks again for your time.

mgallien (changes requested):
@ktec-jp I am sorry for the delay
we value free and open source contributions to the Nextcloud community but we also value the safety and security of the Nextcloud community
as such we do not believe Nextcloud app sending Network ping requests belong to the app store
we think it may be too easy for people to create inadvertent security issues for themselves and other people by installing the app from the app store

From my understanding, that is only part of what the application can do and we would  be happy to still help getting the app published on the appstore but maybe without the ability to support this feature https://github.com/ktec-nc-apps/NetBase#reachability-ping-traceroute-ports-mtu
what do you think ?

ktec-jp:
Thank you for taking the time to look at the application while you are busy — I know the queue is long at the moment.

For what it is worth, I think you are right. I had assumed, too easily, that it was enough for an administrator to be able to switch such a feature off. That is not the same thing as it not being installable from the app store in the first place, and I take the point.

So I have removed the features you raised and rebuilt the application around what is left. Version 0.3.3 drops the reachability tools entirely — ping, traceroute, the TCP port check, TCP ping and path MTU discovery — along with their tabs, their routes and the code behind them. I have removed the nmap front end as well, without waiting to be asked.

What remains is the part the app was really built for: device discovery on the local network, each device's own web interface opened through the server, DNS, whois, TLS and HTTP, subnet and MAC lookups, mail diagnostics, FTP and SFTP, SSH and Telnet, a clock check and the benchmarks. The change is in the repository (https://github.com/ktec-nc-apps/NetBase) and the release notes say plainly why it was made. If anything else in the app gives you the same concern, I would rather hear it now than later.

One more thing, so it does not arrive as a surprise: I am preparing a further app called EditBase, a simple word processor built on HTML. I would be glad to have your eyes on that one too when the time comes.

Thank you again for the offer to help get this published.

mgallien (changes requested):
@ktec-jp thanks for working with us and trying to contribute this new app

can you please fix the failing CN (i.e. common name) check check ?
For more information, have a look at the failing Validate `CSR Common Name` check from the PR

ktec-jp:
Thank you for coming back to me so quickly, and for the care you have taken over this one. You have been patient and courteous through a review that I made longer than it needed to be, and I am grateful for it.

You were right about the common name as well. The CSR I submitted carried a full distinguished name — country, state, organisation and the rest — instead of the common name alone. My three earlier apps were all generated with `CN` by itself; I simply got the command wrong when I created this one, and I am sorry for the extra round trip.

I have regenerated the CSR from the same private key with the subject set to `/CN=netbase`, signed off the commit and force-pushed it (`3eff618`). DCO is green. The `Validate Folder Names` and `Validate CSR Common Name` runs are waiting on a maintainer to approve them — could you start them when you have a moment?

Thank you again, both for the security point you raised earlier and for staying with this through to the end. The app is better for it.

mgallien:
@ktec-jp no worries
you can find the certificate here b6ccf1956c1213a8038b48993c826a509203b82b

PR #1189, opened 2026-08-25
Certificate request for backgroundchanger

chrissi0285:
App ID: `backgroundchanger`

Source code: https://github.com/chrissi0285/backgroundchanger

The CSR is stored at `backgroundchanger/backgroundchanger.csr`; its CN matches the app ID. The private signing key remains outside Git. OpenAI Codex assisted with implementation and testing; maintainer `chrissi0285` is responsible for the app and publication.


mgallien:
@chrissi0285 you can find the certificate here b58c14520a35abf426ae89e6eb0cb5b5c56c37f6

PR #1213, opened 2026-08-30
Add penpot_sync app certificate request

kferrone:
Certificate signing request for the app id **penpot_sync**.

- App id: `penpot_sync`
- Source code (public, AGPL-3.0-or-later): https://github.com/kubed-io/nextcloud-penpot

Penpot Sync mirrors Penpot design files into Nextcloud as native files. Thanks for reviewing.

## Some Lore 

I made this and released it a while back too: https://github.com/kubed-io/nextcloud-n8n
I just wanted to link back to the original that was basically the model for this penpot_sync here. 

Penpot was significantly more challenging to integrate than n8n was. 

Also wanted to point out how I used these two projects as an experiment in how to get quality product from AI using Gherkin as the harness. Oh boy oh boy that worked like a charm.  

Note that I am a platform Engineer, so I left no stone un-turned in the devops department. I'm hoping these projects can be a model for others to make more extensions like this. 

There is even one more still to come for Grafana too! 
https://github.com/kubed-io/nextcloud-grafana

## Fantasy Becomes Reality 

Basically I have always dreamed about Nextcloud (and Google Drive) having perfect integrations with APIs so their data shows up as a file in a file system. GDrive, Nextcloud, Owncloud have always been capable of doing this yet none of them ever ended up with a single "full projection" extension.  What I mean is, a full mirror between the two, each and every move happens in parallel across both apps. 

So I covered pretty much every single move you can think of. Check the `features/` out and you'll see what I mean. 

This was proper challenging and fun and the end result is so so much better than I could have imagined. Enjoy, it's dope yo!


mgallien:
@kferrone you can find the certificate here c4f9364ab7d381d6e7bcc614d109f03946f72846

PR #1227, opened 2026-09-07
Add certificate request for contactduck

Enno-S:
App id contactduck — "ContactDuck", exposes a user's ContactDuck databases as read-only CardDAV address books inside Nextcloud. 

No repository link because the app is built from a private monorepo. This is about my development process, not the licence — the published release is AGPL-3.0 as declared in info.xml, and the tarball carries the complete source of the app.

Questions: info@contactduck.com


mgallien (changes requested):
Can you please include a link to your application repository ?
We do run some quick checks to ensure everything would work when delivering the certificate
See also our documentation here https://nextcloudappstore.readthedocs.io/en/latest/developer.html#obtaining-a-certificate
The rule apply to everybody. That is also regularly needed to early catch errors in the app that would prevent publishing releases on the app store.

mgallien:
gentle ping @Enno-S

PR #1230, opened 2026-09-09
Certificate requests: shortcloud, docsort, filetypes, idregister

CristianCasapu:
Four new apps, all AGPL-3.0-or-later, all mine:

- **shortcloud** — a short link on your own domain for every share (`/go/abc1234`), custom slugs, extra short domains, QR codes, hit counts. https://github.com/CristianCasapu/shortcloud
- **docsort** — reads scanned documents (pdftotext / RapidOCR), recognises what they are and files them in folders by kind. https://github.com/CristianCasapu/docsort
- **filetypes** — which kinds of files each group may upload, decided from the file's bytes and name rather than its extension. https://github.com/CristianCasapu/filetypes
- **idregister** — visitors sign up from a phone with their identity card: live OCR, physical-document check, selfie match. https://github.com/CristianCasapu/idregister

Thanks!


mgallien (changes requested):

mgallien (changes requested):
@CristianCasapu 
I overlooked asking that in my original review.
We need you to split the request into one per certificate. That would mean this one is closed and replaced by 3 new requests.
We need that to track the certificates and their origin.

PR #1246, opened 2026-09-18
Certificate request: source

altipard:
App id: source — an editor for code and configuration files in Nextcloud.


mgallien (commented):

PR #1248, opened 2026-09-18
Rotate certificate request for projectmanager

ricgrangeia:
The private key matching the previously-issued certificate for `projectmanager` was lost, so this replaces the CSR with a new one from a freshly generated keypair. Same app id (`projectmanager`), already published on the App Store (https://apps.nextcloud.com/apps/projectmanager). Source: https://github.com/ricgrangeia/nextcloud-projectmanager


mgallien:
@ricgrangeia you can find the new certificate here f83208dfe735c1cb7df52d7a3e9b28fa7e5b150d

